PR #2, opened 2018-05-25
Fix BaseHtmlElement

lippserd:
Hi Tom,

I fixed some bugs in `BaseHtmlElement` and added PHPDoc where necessary. I added test cases and commit descriptions which should explain the changes. But these are the highlights:

* `render()` did not respect a possible `getDefaultAttributes()` override
* Elements which did not specify a tag were rendered as `<>$content`. It's now just `$content`
* It was allowed to create void elements with content. The content was silently ignored upon rendering. We now throw an Exception here
* `getTag()` overrides were not respected in some places
* `in_array()` is replaced with `isset()` for void detection and the cache has been removed. `setTag()` did not invalidate the cache which may have lead to false results when calling `isVoid()` or `wantsClosingTag()`
* `setVoid()` explicitly supports `null`. Was working before but now it is documented

Cheers,
Eric

### DIFF
--- a/src/BaseHtmlElement.php
+++ b/src/BaseHtmlElement.php
@@ -179,15 +179,4 @@ abstract class BaseHtmlElement extends HtmlDocument
 
         return $this;
     }
-
-    /**
-     * Whether the given something can be rendered
-     *
-     * @param mixed $any
-     * @return bool
-     */
-    protected function canBeRendered($any)
-    {
-        return is_string($any) || is_int($any) || is_null($any);
-    }
 }
--- a/src/BaseHtmlElement.php
+++ b/src/BaseHtmlElement.php
@@ -2,6 +2,8 @@
 
 namespace ipl\Html;
 
+use RuntimeException;
+
 abstract class BaseHtmlElement extends HtmlDocument
 {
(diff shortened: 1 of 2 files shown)


lippserd:
Could we just ignore coding standards for test cases?

lippserd:
One more thing: Should we enforce a tag in `wrap()` or does wrap work without a tag too?

Thomas-Gelf:
I'd strongly suggest to enforce a tag in an HtmlElement, there is no element without a tag in HTML. There is HtmlDocument for everything else, the distinction is intentional. We could of course have one class for "all the things", but that feels weird. How would you render an HtmlElement with attributes but no tag?

lippserd:
Sounds good to me. The only thing which comes to my mind then is to check whether a tag is set in `render()` and `isVoid()` and if not, throw an exception. We can't to that in `getTag()` because it may be overridden.

Thomas-Gelf:
That's fine, please go for it and remove the single commit in question. And don't forget to pull first, I pushed a "fix" for your helper classes.

lippserd:
Affirmative.

Thomas-Gelf (approved):
Nice, thank you!

PR #4, opened 2018-08-06
PHPDoc and small fixes/enhancements for the Attribute and Attributes classes

lippserd:
Most of this PR introduces PHPDoc and some method reordering. Notable fixes are:

* Removed `setName()`. Runtime changes must be forbidden for name. I considered to make this protected but then it is also subject to runtime changes in subclasses.
* `Attribute::render()` renders empty and boolean false attrs as the empty string
* `Attribute::escapeValue()` now expects a glue string
* New method `Attribute::removeValue()`

If you want, I'll squash the PHPDoc related commits once we're finished with the review.

### DIFF
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -21,6 +21,9 @@ class Attribute
     /** @var string|array|bool|null */
     protected $value;
 
+    /** @var string Glue string to join elements if the attribute's value is an array */
+    protected $glue = ' ';
+
     /**
      * Create a new HTML attribute from the given name and value
      *
@@ -211,7 +214,7 @@ class Attribute
      */
     public function renderValue()
     {
-        return static::escapeValue($this->value);
+        return static::escapeValue($this->value, $this->glue);
     }
 
     /**
@@ -225,14 +228,23 @@ class Attribute
     }
 
     /**
-     * @param $value
-     * @return string
+     * Escape the value of an attribute
+     *
+     * If the value is an array, returns the string representation
+     * of all array elements joined with the specified glue string.
+     *
+     * Escaping takes place using {@link Html::encode()}.
+     *
+     * @param   string|array    $value
+     * @param   string          $glue   Glue string to join elements if value is an array
+     *
+     * @return  string
      */
-    public static function escapeValue($value)
+    public static function escapeValue($value, $glue = ' ')
     {
         // TODO: escape differently
         if (is_array($value)) {
-            return Html::escape(implode(' ', $value));
+            return Html::escape(implode($glue, $value));
         } else {
             return Html::escape((string) $value);
         }
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -79,7 +79,9 @@ class Attribute
     }
 
     /**
-     * @return string
+     * Get the value of the attribute
+     *
+     * @return  string|bool|array|null
      */
     public function getValue()
     {
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -142,7 +142,9 @@ class Attribute
     }
 
     /**
-     * @return bool
+     * Test and return true if the attribute is boolean, false otherwise
+     *
+     * @return  bool
      */
     public function isBoolean()
     {
@@ -150,7 +152,11 @@ class Attribute
     }
 
     /**
-     * @return bool
+     * Test and return true if the attribute is empty, false otherwise
+     *
+     * Null and the empty array will be considered empty.
+     *
+     * @return  bool
      */
     public function isEmpty()
     {
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -150,6 +150,14 @@ class Attribute
     }
 
     /**
+     * @return bool
+     */
+    public function isEmpty()
+    {
+        return null === $this->value || $this->value === [];
+    }
+
+    /**
      * Render the attribute to HTML
      *
      * If the value of the attribute is of type boolean, it will be rendered as
@@ -198,14 +206,6 @@ class Attribute
     public function renderValue()
     {
         return static::escapeValue($this->value);
-    }
-
-    /**
-     * @return bool
-     */
-    public function isEmpty()
-    {
-        return null === $this->value || $this->value === [];
     }
 
     /**
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -35,10 +35,14 @@ class Attribute
     }
 
     /**
-     * @param $name
-     * @param $value
-     * @return static
-     * @throws InvalidArgumentException
+     * Create a new HTML attribute from the given name and value
+     *
+     * @param   string                  $name   The name of the attribute
+     * @param   string|bool|array|null  $value  The value of the attribute
+     *
+     * @return  static
+     *
+     * @throws  InvalidArgumentException        If the name of the attribute contains special characters
      */
     public static function create($name, $value)
     {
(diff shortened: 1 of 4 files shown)


src/Attribute.php
```diff
@@ -130,6 +131,36 @@ public function addValue($value)
...
+    /**
+     * Remove the attribute's value or remove the given value from the attribute
+     *
+     * @param   string|string[] $value      The value to remove
+     *
+     * @return  $this
+     *
+     * @throws  UnexpectedValueException    If given the value to remove and the current value is not an array
+     */
+    public function removeValue($value = null)
```
Thomas-Gelf: Please make $value required, IMO who wants to set an attribute to `null` without removing the Attribute should be forced to run `setValue(null)`.

src/Attribute.php
```diff
@@ -130,6 +131,36 @@ public function addValue($value)
...
+    {
+        if ($value === null) {
+            $this->value = null;
+
+            return $this;
+        }
+
+        if (! is_array($this->value)) {
```
Thomas-Gelf: Please relax this. It is perfectly legal to have an `Attribute` with a string value and adding another value later on. You do not have to care whether it is already an array or not, same should apply when removing values. If the current value is a string and the string matches the given value it should be nulled, if it is a string which doesn't match it should do nothing and return without an error.

src/Attributes.php
```diff
@@ -93,31 +93,54 @@ public function get($name)
     }
 
     /**
```
Thomas-Gelf: Just a generic note -> please do not push commits which change and move code in one commit - it's extremely annoying to read them, and you have to be very, very careful to not overlook changes.

src/Attributes.php
```diff
@@ -93,31 +93,54 @@ public function get($name)
...
+     *
+     * @throws  InvalidArgumentException    If the attribute does not yet exist and its name contains special characters
      */
     public function add($attribute, $value = null)
     {
-        // TODO: do not allow Attribute and Attributes
-        if ($attribute instanceof Attributes) {
-            foreach ($attribute->getAttributes() as $a) {
-                $this->add($a);
+        if ($attribute instanceof static) {
```
Thomas-Gelf: `Attributes` is correct here, `static` is wrong. Reasoning: when extending this class you still want "normal" `Attributes` to be accepted, wouldn't you? Same goes true for `get`, there it has already been wrong. You can correct that one too if you want.

src/Attribute.php
```diff
@@ -208,12 +208,16 @@ public function renderValue()
...
+     *
+     * Makes sure that the name of an attribute really is a string.
+     *
+     * @param   string  $name
+     *
+     * @return  string
      */
     public static function escapeName($name)
     {
-        // TODO: escape
```
Thomas-Gelf: Please leave the TODO in place unless resolved.

src/Attribute.php
```diff
@@ -211,14 +214,22 @@ public static function escapeName($name)
...
+     *
+     * @param   string|array    $value
+     * @param   string          $glue   Glue string to join elements if value is an array
+     *
+     * @return  string
      */
-    public static function escapeValue($value)
+    public static function escapeValue($value, $glue)
     {
-        // TODO: escape differently
```
Thomas-Gelf: One more unresolved TODO, should remain in place.

src/Attribute.php
```diff
@@ -211,14 +214,22 @@ public static function escapeName($name)
...
+     *
+     * Escaping takes place using {@link Html::encode()}.
+     *
+     * @param   string|array    $value
+     * @param   string          $glue   Glue string to join elements if value is an array
+     *
+     * @return  string
      */
-    public static function escapeValue($value)
+    public static function escapeValue($value, $glue)
```
Thomas-Gelf: Can we leave `$glue = ' '` as the default here?

src/Attribute.php
```diff
@@ -52,24 +61,6 @@ public function getName()
         return $this->name;
     }
 
-    /**
```
Thomas-Gelf: Could you please keep `setName()` and make it protected? I hate constructors with too much logic, it's always a mess when you want to extend/override them.

src/Attribute.php
```diff
@@ -120,8 +120,16 @@ public function isBoolean()
...
     {
-        if ($this->isBoolean() && $this->value) {
-            return $this->renderName();
+        if ($this->isEmpty()) {
+            return '';
+        }
+
+        if ($this->isBoolean()) {
+            if ($this->value) {
+                return $this->renderName();
```
Thomas-Gelf: Given these options:

```html
<input required />
<input required="required" />
<input required="" />
```
Do you have any preference?

Thomas-Gelf (changes requested):

Thomas-Gelf:
Thanks for this, I added some comments

src/Attribute.php
lippserd: 👍 I considered this but decided to just remove it so it's not that easy to change the name in subclasses. But since it's possible to override the constructor anyways, I'll make it protected.

src/Attribute.php
lippserd: The first one.  We render HTML5.

src/Attribute.php
lippserd: Sure!

src/Attribute.php
lippserd: What should be escaped differently here?

src/Attribute.php
lippserd: There is no escape for attribute names. Either it is valid or invalid.

src/Attributes.php
lippserd: Absolutely. I'll split the commits in question.

src/Attribute.php
Thomas-Gelf: That's what I would also have said. However:

Thomas-Gelf: [3.2.2 Boolean attributes](https://www.w3.org/TR/2008/WD-html5-20080610/semantics.html):
> If the attribute is present, its value must either be the empty string or a value that is a case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.

src/Attribute.php
Thomas-Gelf: The attribute value should not be escaped like any HTML text node, as [attributes](https://www.w3.org/TR/html5/syntax.html#elements-attributes) have different semantics. This implementation should be safe, but it is not correct.

Thomas-Gelf: Googled a nice [example](https://mathiasbynens.be/notes/html5-id-class):

```html
<p id="#">Foo.
<p id="##">Bar.
<p id="♥">Baz.
<p id="©">Inga.
<p id="{}">Lorem.
<p id="“‘’”">Ipsum.
<p id="⌘⌥">Dolor.
<p id="{}">Sit.
<p id="[attr=value]">Amet.
<p id="++++++++++[>+++++++>++++++++++>+++>+<<<<-]>++.>+.+++++++..+++.>++.<<+++++++++++++++.>.+++.------.--------.>+.>.">Hello world!
```

I want to be able to use those.

src/Attribute.php
lippserd: You are looking at a working draft 😆 

Please see here: https://www.w3.org/TR/html5/infrastructure.html#sec-boolean-attributes

src/Attribute.php
Thomas-Gelf: You're right, we should drop the method and replace it with `assertValidName` eventually throwing an `InvalidArgumentException`.

src/Attributes.php
Thomas-Gelf: Don't worry, it's not worth the effort. Next time!

src/Attribute.php
Thomas-Gelf: Thanks Google. I knew my code must have been correct :laughing:

Btw:

>  A boolean attribute without a value assigned to it (e.g. checked) is implicitly equivalent to one that has the empty string assigned to it (i.e. checked="").

I was also wondering because Firefox Developer tools continue to show `<input required="" />` while our code says `<input required />`. The one we opted for looks better to me, let's keep it.

lippserd:
Made all the changes you've requested. I'd like to postpone the name validation in order to get this merged.

Thomas-Gelf (approved):

PR #6, opened 2018-10-31
WIP: Feature/phrasing content

theFeu:
Needs review and squash

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -12,7 +12,7 @@
     >
     <testsuites>
         <testsuite name="ipl PHP Unit tests">
-            <directory suffix=".php">tests/php</directory>
+            <directory suffix="Test.php">tests/php</directory>
         </testsuite>
     </testsuites>
 </phpunit>
--- a/src/BaseHtmlElement.php
+++ b/src/BaseHtmlElement.php
@@ -48,6 +48,8 @@ abstract class BaseHtmlElement extends HtmlDocument
      * If {@link $contentSeparator} is null, this property should be used to decide whether there should be a newline or
      * no content separator rendered.
      *
+     * We added the <option> tag.
+     *
      * @var array
      *
      * @see https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content
@@ -80,6 +82,7 @@ abstract class BaseHtmlElement extends HtmlDocument
         'meter'     => 1,
         'noscript'  => 1,
         'object'    => 1,
+        'option'    => 1,
         'output'    => 1,
         'progress'  => 1,
         'q'         => 1,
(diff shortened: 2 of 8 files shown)


Thomas-Gelf:
@theFeu: I'm wondering about the motivation for this pull request. I'd definitively prefer
```html
<h1>Hello <strong>there</strong>, are your okay?</h1>
```
over 
```html
<h1>
Hello 
<strong>there</strong>
, are your okay?
</h1>
```

Could you please share your related thoughts to get a better understanding of the problem you're trying to tackle?

Thanks,
Thomas

theFeu:
> @theFeu: I'm wondering about the motivation for this pull request. I'd definitively prefer
> 
> ```
> <h1>Hello <strong>there</strong>, are your okay?</h1>
> ```
> 
> over
> 
> ```
> <h1>
> Hello 
> <strong>there</strong>
> , are your okay?
> </h1>
> ```
> 
> Could you please share your related thoughts to get a better understanding of the problem you're trying to tackle?
> 
> Thanks,
> Thomas

The thought behind it was, that the ```h1```-tag is not to be treated as phrasing content. ```h1```is therefore to be treated like a ```div```, in which all content is to be put on a new line.

I do agree that it looks silly, but it's got to for conformity :)

This page treats the issue similarly: 
```
<h1 class="gh-header-title">
 <span class="js-issue-title">
 WIP: Feature/phrasing content
 </span>
 <span class="gh-header-number">#6</span>
</h1>
```

Please do correct me if that doesn't really look right to you and thanks for the feedback!
Feu

Thomas-Gelf:
Talked offline to @lippserd and @theFeu. Main concern is that such a basic library should not show unexpected behavior. Given this example:

```html
<style>
.tile {
  height: 10em;
  width: 10em;
  display: inline-bock;
}
</style>
<div><div class="tile" /><div class="tile" /></div>
``` 
How should we know whether the library consumer expects to see tiles separated by spaces?

We could use the strategy pattern to implement this:
```php
HtmlDocument->setSeparatorStrategy(new W3cBestPracticeSeparatorStrategy());
```
At rendering time the `HtmlDocument` would tell it's children about it's `$separatorStrategy` if any. They will then:
* respect it unless they have a custom `$contentSeparator` or `$separatorStrategy`
* hand it over to their children unless they have a custom `$separatorStrategy`

We could ship a `W3C` (strict) and a `IcingaWeb2` (relaxed) strategy. `Relaxed` could for example use newlines for `div`s and `ul`s but not for `h*` and `li`.

Cheers,
Thomas

Thomas-Gelf:
@lippserd: how should we proceed here?

lippserd:
> @lippserd: how should we proceed here?

Since we never really had the need to implement this properly, I think it's perfectly fine to close this topic.

PR #11, opened 2019-01-21
Support multiple submit buttons

lippserd:
This PR introduces new methods to support multiple submit buttons.
It does not interfere with the idea that our forms have one primary submit button.

The primary submit button is handled via `has/get/setSubmittButton()` and `hasBeenSubmitted()`.
Other pressed submit buttons are retrieved via `getPressedSubmitElement()`.

The question here is how we handle form redirects:

For the primary submit button we call `hasBeenSubmitted() and isValid()`.

For other submit buttons we have to do something like the following:
`getPressedSubmitElement() && (! getPressedSubmitElement()->getAttributes()->get('formnovalidate') || isValid())`.

@Thomas-Gelf I'd like to provide this as method in our forms. Maybe something like `isComplete()`?

We also talked about onsuccess callbacks and the method onSend. I'd like to address this in a separate PR.

### DIFF
--- a/src/Contract/FormSubmitElement.php
+++ b/src/Contract/FormSubmitElement.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace ipl\Html\Contract;
+
+interface FormSubmitElement
+{
+    /**
+     * Get whether the element has been pressed
+     *
+     * @return bool
+     */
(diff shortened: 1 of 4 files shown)


lippserd:
Superseded by #27

PR #12, opened 2019-01-21
Standardize tests

lippserd:
@Thomas-Gelf Any objections against this one? These are the same changes as in ipl\stdlib. I'd like to standardize the testing stuff in all our ipl repos.


Thomas-Gelf (approved):
Looks good to me!

PR #13, opened 2019-01-22
[WIP] Fix supported validator specs

lippserd:
With this PR we support the following methods for setting validators:

```php
// Name to options
$validators = [
    'strlen' => [
        'min' => 8,
         'max' => 32
    ]
];

// Array
$validators = [
    [
        'name' => strlen,
        'options' => ...
    ],
    [
        'name' => 'email'
    ]
];

// Simple
$validators = ['email', 'iban'];

// Instance of ValidatorInterface
$validators = [new EmailValidator(), ...];
```

Before, we just supported the name to options spec.

@Thomas-Gelf I'd like to move the validator factory functionality from `setValidators()` to a dedicated class/function, e.g. `Validators` or `ValidatorChain` with `wantValidators`. There we may add something like Zend's break on failure and registering/loading of custom validators via our stdlib. What do you think? `getValidators()` in the form element would then return an instance of this class.

### DIFF
--- a/src/FormElement/BaseFormElement.php
+++ b/src/FormElement/BaseFormElement.php
@@ -239,7 +239,7 @@ abstract class BaseFormElement extends BaseHtmlElement
      * @param $options
      * @return ValidatorInterface
      */
-    public function createValidator($name, $options)
+    public function createValidator($name, $options = null)
     {
         $class = 'ipl\\Validator\\' . ucfirst($name) . 'Validator';
         if (class_exists($class)) {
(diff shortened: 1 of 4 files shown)


Thomas-Gelf:
Hint: this is waiting for a ValidatorChain. Also, please consider using `PluginLoader`.

lippserd:
This is no longer required. https://github.com/Icinga/ipl-validator now ships common validators and the validator chain.

PR #14, opened 2019-01-28
Html: introduce ::wrapEach()

Thomas-Gelf:


src/Html.php
```diff
@@ -85,6 +86,42 @@ public static function sprintf($string)
...
+     * a given callback/callable is being called passing key and value of each
+     * list entry as parameters
+     *
+     * @param array|Traversable $list
+     * @param string|callable $wrapper
+     * @return HtmlDocument
+     */
+    public static function wrapEach($list, $wrapper)
+    {
+        if (! is_array($list) && !($list instanceof Traversable)) {
```
lippserd: You may remove the parentheses around the instanceof check.

lippserd (changes requested):

src/Html.php
Thomas-Gelf: thnx!

lippserd (approved):

lippserd:
You may merge once you're ready 👍

PR #20, opened 2019-05-16
WIP: Introduce FormElementInterface

Thomas-Gelf:
@lippserd: I do not like this interface, it is huge. We might want to combine multiple interfaces, like `MessageContainerInterface` and more. Decorator, Form and others might then apply different logic depending on the available feature-set.


lippserd:
Superseded by #27

PR #22, opened 2019-10-01
HtmlString: Escape content if instructed to do so

nilmerg:


nilmerg:
*Whoever* wrote the documentation then, interpreted this just like I did. (https://github.com/Icinga/ipl-html/blob/master/doc/draft.md#text-usage and https://github.com/Icinga/ipl-html/blob/master/doc/draft.md#htmlstring)

You're right that the property should read `escape` then. Though, just as I just showed with this PR that I didn't noticed this others won't as well I guess. I mean, there are two classes. One for text (in the sense of *human readable*) and the other for raw HTML. That's `Text` and `HtmlString`.

If you however insist on the property name, may we switch the class names then?

Thomas-Gelf:
Those two classes address two distinct use-cases:

* `Text`: give me random text, I'll make sure that everything will be safe
* `HtmlString`: shut up and take my raw HTML, I know what I'm doing

You should use `HtmlString` only when sure that your string is valid HTML This is useful when using self-rendering third-party components. The `escaped` property should work accordingly:

* `Text::$escaped = false`: my content hasn't been escaped, please take care
* `HtmlString::$escaped = true` what I provided has been escaped, trust me

To me it seems that this is exactly what happens right now, and the documentation should reflect this. Eventually I'm overlooking something. If so, please let's have a short chat to clear things up - that might be more effective :-)

Cheers,
Thomas

nilmerg:
I'm totally with you. Though, it's not the case. That's the reason for this PR. :wink: 

Just try it for yourself: `echo (new HtmlString('<hr>'))->render();` == `"&lt;hr&gt;"` (on master of course)

PR #23, opened 2019-10-16
More strict interfaces

nilmerg:


src/FormDecorator/DivDecorator.php
```diff
@@ -109,10 +109,11 @@ protected function assemble()
 
         $this->formElement->getAttributes()->add('class');
 
-        $this->addHtml(...array_filter(array_merge([
```
nilmerg: I used `array_merge` here because `assembleErrors` returns an array. All other methods don't. Without it, `addHtml` gets an array instead of `ValidHtml`.

src/Html.php
```diff
@@ -164,18 +158,20 @@ public static function wantHtml($any)
...
+        $list = [];
+
         if ($content === null) {
             return $list;
-        } elseif (! is_array($content) && ! $content instanceof Traversable) {
+        } elseif (! is_iterable($content)) {
             $list[] = static::wantHtml($content);
         } else {
-            foreach ($content as $entry) {
-                static::wantHtmlList($entry, $list);
```
nilmerg: Is using `array_merge` here faster than adjusting a reference with each call?

src/Html.php
```diff
@@ -164,18 +158,20 @@ public static function wantHtml($any)
...
      */
-    public static function wantHtmlList($content, &$list = [])
+    public static function wantHtmlList($content)
     {
+        $list = [];
+
         if ($content === null) {
             return $list;
-        } elseif (! is_array($content) && ! $content instanceof Traversable) {
+        } elseif (! is_iterable($content)) {
```
nilmerg: `is_iterable` is php 7.1+ only.

src/Html.php
```diff
@@ -28,23 +28,18 @@ abstract class Html
      */
     public static function tag($name, $attributes = null, $content = null)
     {
-        if ($content !== null) {
-            $content = static::wantHtmlList($content);
-        } elseif ($attributes instanceof ValidHtml || is_scalar($attributes)) {
-            $content = static::wantHtmlList($attributes);
+        if ($attributes instanceof ValidHtml || is_scalar($attributes)) {
```
nilmerg: I just noticed why I changed the implementation. The intention was to focus on the most probable path calls to this factory method will follow.

That's: `Html::tag('tag', ['attributes'], Html::tag('content'))`

In this case the current implementation still checks if `$attributes` may be `$content` 100% of the time.

nilmerg:
Superseded by #51

PR #25, opened 2020-02-06
HtmlDocument::addFrom()

nilmerg:
Example usage:

```php
$columns->addFrom($hostStats, function (BaseHtmlElement $item) {
    $item->getAttributes()->add(['class' => 'col']);
    $item->setTag('div');
    return $item;
});
```

### DIFF
--- a/doc/10-Quickstart.md
+++ b/doc/10-Quickstart.md
@@ -237,3 +237,21 @@ Still, we do not consider this code being very readable. In case you ever used
 ```html
 Hi <strong>there</strong>, are you ok?
 ```
+
+### Passing Elements
+
+Say you've now written your own wonderful widget. It's even so wonderful you want to re-use some of its components
+in another widget? You're in luck, there's a handy shortcut with which you can easily pass elements from one widget
+to another.
+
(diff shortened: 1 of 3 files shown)


Thomas-Gelf:
@nilmerg: `Html::wrapEach()` could be what you're looking for:

```php
$columns->add(Html::wrapEach($hostStats, function ($key, $value) {
    // ...
}));
```

nilmerg:
Nope, I don't want another document layer. I simply need an approach to get items from one container and add them to another one while being able to transform them in the process.

e.g. this (the ul is `$hostStats`)
```html
<ul>
    <li>...</li>
    <li>...</li>
</ul>
```

to this: (the div is `$columns`)
```html
<div>
      <div>...</div>
      <div>...</div>
</div>
```

`wrapEach` doesn't allow me to do this in the same manner as `addFrom` does it. It also feels hacky using it for the intended effect.

lippserd:
I'd like to see tests and maybe a section about this method in our docs.
And I'm not quite sure whether PHP optimizes the code but since we're trying to be as performant as possible with this lib, the check for the callback should be outside of the foreach.

nilmerg:
Done.

PR #32, opened 2020-05-27
FormattedString: Support sprintf-like behavior for non string arguments

lippserd:


src/FormattedString.php
```diff
@@ -51,7 +51,15 @@ public function __construct($format, $args = null)
...
             foreach ($args as $key => $val) {
-                $this->args[$key] = Html::wantHtml($val);
+                if (! $val instanceof ValidHtml) {
+                    if (
+                        is_string($val)
+                        || is_object($val) && method_exists($val, '__toString')
+                    ) {
+                        $val = new Text($val);
+                    }
+                }
```
nilmerg, line 61: Currently this ignores arrays and fails. Why not invert the logic and still use `Html::wantHtml()`?

```suggestion
                if (! is_scalar($val) || is_string($val)) {
                    $val = Html::wantHtml($val);
                }
```

lippserd: Makes sense, ty!

PR #34, opened 2020-10-15
New form event to handle autosubmits

nilmerg:
This adds the event `ON_SENT` to class `Form` which is emitted once a form has been partially validated.

This allows to register handling for forms which is run in any case once it's submitted.
Though, actions to run upon user submits are still possible in `ON_SUCCESS` events.

The added property for the redirection url is a complement to this, as it allows to define the url in `ON_SENT` and then use it either in `ON_SUCCESS` or custom code that runs later checking for `Form::hasBeenSent()`.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -20,6 +20,7 @@ class Form extends BaseHtmlElement
     const ON_ERROR = 'error';
     const ON_REQUEST = 'request';
     const ON_SUCCESS = 'success';
+    const ON_SENT = 'sent';
 
     /** @var string Form submission URL */
     protected $action;
@@ -220,6 +221,7 @@ class Form extends BaseHtmlElement
         if (! $this->hasSubmitButton() || $this->getSubmitButton()->hasBeenPressed()) {
             if ($this->isValid()) {
                 try {
+                    $this->emit(Form::ON_SENT, [$this]);
                     $this->onSuccess();
                     $this->emitOnce(Form::ON_SUCCESS, [$this]);
                 } catch (Exception $e) {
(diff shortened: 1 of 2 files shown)


src/Form.php
```diff
@@ -7,6 +7,7 @@
 use ipl\Html\Contract\FormSubmitElement;
 use ipl\Html\FormElement\FormElements;
 use ipl\Stdlib\Messages;
+use ipl\Web\Url;
```
Thomas-Gelf, line 10: ipl\Html shouldn't depend on ipl\Web

Thomas-Gelf (changes requested):

src/Form.php
nilmerg: You're right. Needs to be string then I guess

Thomas-Gelf: Thanks!

src/Form.php
```diff
@@ -155,6 +159,30 @@ public function setRequest($request)
...
+    /**
+     * Set the url to redirect to on success
+     *
+     * @param string $url
+     *
+     * @return $this
+     */
+    public function setRedirectUrl($url)
+    {
+        $this->redirectUrl = (string) $url;
```
lippserd, line 181: Any objection to not using the string cast here? If we put an URL object in `setRedirectUrl()` it shouldn't be changed in my opinion.

nilmerg: That's a matter of interface stability imho. If this can be a url object, or any other type, anyone using the getter must first check the return value's type which shouldn't be necessary for a getter. I'm also not really a fan of this being a string only, but if we have no url object outside ipl-web this is the only solution for me.

nilmerg: Well, on a second thought.. If anyone using the getter handles the return value as string and the actual type allows implicit string casting.. Removing the explicit cast here is fine, I think.

lippserd (approved):

PR #35, opened 2020-11-06
Form: Validate empty elements also in `validatePartial()`

nilmerg:
Empty elements may also be valid.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -325,9 +325,7 @@ class Form extends BaseHtmlElement
     public function validatePartial()
     {
         foreach ($this->getElements() as $element) {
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
@lippserd: are you sure about this? `validatePartial()` is there to validate only what the user already compiled. Without that `if` statement the method provides no additional value. A user does not expect to get all remaining required fields red when toggling an auto-submit element, because he did not yet express the intent to submit the form.

lippserd:
Only `Form::validate()` checks for the `required` attribute. This one is about the programmatically `valid` state.

Thomas-Gelf:
Thanks!

PR #37, opened 2021-02-19
Make tests compatible with PHPUnit 9

lippserd:


nilmerg:
You should probably also enable PHPUnit 9 then in the actions workflow.

nilmerg:
Ooops, sorry, forgot this and changed it already with #44.

PR #41, opened 2021-04-23
Support mustache template mapping

### LINKED ISSUE #33: Templating to aid enriching localized content with HTML
## Is your feature request related to a problem? Please describe.

Content that can be translated to other languages may sometimes require changes to its structure. `Total %d` for example may get translated in German to `%d Gesamt`.
There may also be entire sentences which get translated. If they're translated to a language such as Arabic the sentence may reversed as textual content in Arabic is written from right to left (RTL).

To make this possible, source code must mark the entire content in a single message as localizable. e.g. `t('Total %d')`. String concatenation is not possible.

In case a developer wants to enrich parts of the content with HTML, this is currently not possible in such a case. Say I'd want to put `Total` in a span to style it differently, the result would look like this: (Example taken from https://github.com/Icinga/icingadb-web/pull/94/files)

`<span class="meta">Total</span> %d`

Though, this would require rendering the HTML instantly and allow the translator to change the markup. Both are undesirable side-effects.

## Describe the solution you'd like

There is already the possibility to format content using the class `FormattedString`. Why not extend it so that it supports a simple form of templating. This could look like this:

`{{#total}}Total{{/total}} %d`

Given the following code:

```php
FormattedString::create(
    '{{#total}}Total{{/total}} %d',
    ['total' => Html::tag('span')]
);
```

The result could be:

`<span>Total</span> 1234`

and in German:

`1234 <span>Gesamt</span>`

raviks789:
Earlier FormattedString class did not support mustache template mapping. Here mustache template mapping
for Html tags has been introduced.

fixes #33


Thomas-Gelf (changes requested):
Please note that the solely purpose of the `FormattedString` class is to support `formatted strings` as known from the C language. While personally not feeling the immediate need for such a solution, I have no objections against introducing support for a templating language. However, I'd suggest moving it to a dedicated class, IHMO `FormattedString` is the wrong place for this.

Thomas-Gelf (approved):

lippserd:
I agree that this should be something like `TemplatedString`.

nilmerg (changes requested):
We had the discussion about the possible interfaces:

* `...('{{#total}}Total{{/total}} %d', ['total' => Html::tag('span')], 42)`
* `...('{{#total}}Total{{/total}} %d', total: Html::tag('span'), 42)` (PHP8+)
* `...('{{#total}}Total{{/total}} %d', 42)->setMustacheTemplates(['total' => Html::tag('span')])`

The latter is what we agreed upon eventually. The second one would be the cleanest, but it's no option. The first however, although I said it's not the best, is still the best compromise I think. (Now, when thinking about it again)

It should be possible to support this:

`$title = TemplateString::create('{{#total}}Total{{/total}} %d {{#num}}#{{/num}}', ['total' => Html::tag('span'), 'num' => Html::tag('i')], 42);`

The position of the template arguments though is **not** important. Because I think of the processing this way:

1) Filter the arguments by type (scalar values are parameters for `vsprintf`, arrays for the template)
2) `vsprintf` arguments are still registered in their respective order so that when passing them the result is this: `'{{#total}}Total{{/total}} 42 {{#num}}#{{/num}}'`
3) arrays are merged together so that at the end a single map exists to fill the template with: `'<span>Total</span> 42 <i>#</i>'`

So please change the implementation so that it works this way.
Also, please do the following:

* Rename the class to `TemplateString` (@lippserd *Templated* isn't a word :teacher: )
* Let the class inherit from `FormattedString`, and let it perform the `vsprintf` handling
  * For this to work, you'll need to filter the arguments by type in the constructor
* The template processing should happen deferred, not instantly
  * Perform it as part of method `render()` instead

I'll provide you next week with additional test cases of the various usages I imagine for this.

nilmerg:
I pushed some additional test cases now. Though, to a different branch to not disrupt your workflow. Once pulled, you can cherry pick it instead: `git cherry-pick 977fbff`

src/TemplateString.php
```diff
@@ -0,0 +1,230 @@
...
+    protected $templateArgs;
+
+    protected $pos;
+
+    /** @var int Length of {{#}} | {{/}}*/
+    const MUSTACHE_LENGTH = 5;
+
+    public function __construct($format, $args = null)
+    {
+        $this->templateArgs = [];
```
nilmerg, line 34: You could already initialize this with an empty array. (In the property definition)

src/TemplateString.php
```diff
@@ -0,0 +1,230 @@
...
+
+        foreach ($args as $key => $val) {
+            if (is_array($val) && $this->isAssoc($val)) {
```
nilmerg, line 37: I believe you can replace `$this->isAssoc($val)` with `is_string(key($val))`. If the first key isn't a string the array isn't a template argument. If there are indeed other keys which are not strings, the rendering should fail in my opinion. String keys don't matter for `FormattedString` anyway, so a developer shouldn't mix key types in any case. And if it's done, it should throw an error instead of silently accepting it.

src/TemplateString.php
```diff
@@ -0,0 +1,230 @@
...
+                unset($args[$key]);
+                $this->templateArgs[$key] = $val;
```
nilmerg, line 39: If you'd use `$this->templateArgs += $val` you can remove the array_merge below. The `+` operator with arrays doesn't override existing keys though, but I don't think that's an issue. If we wouldn't have to support PHP 7+ ( :rofl: ) we could use PHP 8's named args, then arg names would be unique anyway.

src/TemplateString.php
```diff
@@ -0,0 +1,230 @@
...
+            }
+        }
+
+        if (!empty($this->templateArgs)) {
+            $this->templateArgs = call_user_func_array('array_merge', $this->templateArgs);
+        }
+
+        $args = array_values($args);
```
nilmerg, line 47: Please already populate an array during the loop. It's more efficient that way.

src/TemplateString.php
```diff
@@ -0,0 +1,230 @@
...
+
+    public function render()
+    {
+        $formattedstring = parent::render();
+        if (empty($this->templateArgs)) {
+            return $formattedstring;
+        }
+
+        $formattedstring = $this->parseTemplates($formattedstring)->render();
+        return $formattedstring;
```
nilmerg, line 228: That could be a single line :wink:

nilmerg (changes requested):
Left some comments in the constructor and the `render` method. Regarding the actual parsing, we should talk directly. It's too difficult to give my thoughts based on text :sweat_smile:

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+     * @param null $for
+     * @return HtmlDocument
+     * @throws Exception in case of missing template argument or unbounded open or close templates
+     */
+    protected function parseTemplates($for = null)
+    {
+        $buffer = '';
+
+        while ($this->pos < $this->length) {
+            $char = $this->readChar();
```
nilmerg, line 59: You could combine that like so:

```suggestion
        while (($char = $this->readChar()) !== false) {
```

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+                } else {
+                    // substrings of form {{#foo, {{/foo {{#foo} or {{/foo} treated as a normal string
+                    $strlen = strlen($templateKey);
+                    $this->pos -= ($strlen + 2);
+
+                    $buffer .= $char;
+                }
+            } else {
+                $buffer .= $char;
+            }
```
nilmerg, line 116: You could move that to the top (`if ($char !== '{') {` and combine it with a `continue`. This way you can remove one level of indentation and what happens for non-template chars is instantly obvious.

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        if ($this->pos === $this->length && !empty($this->unclosed)) {
+            throw new Exception(vsprintf(
+                'Unbounded opening of template: %s',
+                $this->unclosed
+            ));
+        }
+        $buffer = HtmlString::create($buffer);
+        return (new HtmlDocument())->add($buffer);
```
nilmerg, line 126: Please combine both lines, and put an empty line before them. There's a control structure beforehand after all.

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+     */
+    protected function parseTemplates($for = null)
+    {
+        $buffer = '';
+
+        while ($this->pos < $this->length) {
+            $char = $this->readChar();
+
+            if ($char === '{') {
+                $this->pos++;
```
nilmerg, line 62: What if there's not a `{` after? Please check for that as well.

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+                $start = $this->readChar() === '#';
+                $this->pos -= 1;
+
+                $end = $this->readChar() === '/';
```
nilmerg, line 66: I'm sure that's solvable with a single call to `readChar` ;)

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+
+                $templateKey = $this->readUntil('}');
+                // if the string following '{{#' is read up to the last character or (length - 1)th character
+                // then it is not a template
+                if ($this->pos >= $this->length - 1) {
+                    $strlen = strlen($templateKey);
+                    $this->pos -= ($strlen + 3);
+                    $buffer .= $this->readUntil();
```
nilmerg, line 74: Can't you just do this?: (`$templateHandle` is what I imagine being the result of the single call to `readChar` above :wink: )

```suggestion
                    $buffer = '{{' . $templateHandle . $templateKey;
```

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+                    throw new Exception(sprintf(
+                        'Unbounded closing of template: %s',
+                        $templateKey
+                    ));
+                } else {
+                    // substrings of form {{#foo, {{/foo {{#foo} or {{/foo} treated as a normal string
+                    $strlen = strlen($templateKey);
+                    $this->pos -= ($strlen + 2);
+
+                    $buffer .= $char;
```
nilmerg, line 112: With this, you're essentially parsing the chars all over again. Please do it the same as I suggested for lines 72 to 74.

src/TemplateString.php
```diff
@@ -0,0 +1,178 @@
...
+                // then it is not a template
+                if ($this->pos >= $this->length - 1) {
+                    $strlen = strlen($templateKey);
+                    $this->pos -= ($strlen + 3);
+                    $buffer .= $this->readUntil();
+                    break;
+                }
+
+                $this->pos++;
+                $next = $this->string[$this->pos--] === '}';
```
nilmerg, line 79: Please take a look here again. Why are you post-decrementing the position? Why are you incrementing it by `2` in various places below? There's too much forward and backward to me here. The increments by `2` below shouldn't be necessary as I see it.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+            $templateKey = $this->readUntil('}');
+            // if the string following '{{#' is read up to the last character or (length - 1)th character
+            // then it is not a template
+            if ($this->pos >= $this->length - 1) {
+                $buffer .= $char . $templateHandle . $templateKey;
+                break;
+            }
+
+            $this->pos++;
+            $next = $this->string[$this->pos] === '}';
```
nilmerg, line 81: Please use `readChar` here. You're interested in the next char and later on you're also incrementing the position. The method does both things already.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+            } elseif ($end && $next) {
+                // throw exception for unbounded closing of templates
+                throw new Exception(sprintf(
+                    'Unbounded closing of template: %s',
+                    $templateKey
+                ));
+            } else {
+                // substrings of form {{#foo, {{/foo {{#foo} or {{/foo} treated as a normal string
+                $buffer .= $char . $templateHandle . $templateKey;
+            }
```
nilmerg, line 112: Please use that as the first `if`. Then you can remove the `&& $next` from all other branches. (I'd request to add that to the `$for === $templateKey` condition anyway.) The `if` should check directly for `$nextChar !== '{'` and use `$nextChar` to populate the buffer.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+                    $wrapper = $this->templateArgs[$templateKey];
+
+                    $buffer .= $this->parseTemplates($templateKey)->prependWrapper($wrapper);
+                } else {
+                    throw new Exception(sprintf(
+                        'Missing template argument: %s ',
+                        $templateKey
+                    ));
+                }
+            } elseif ($for === $templateKey) {
```
nilmerg, line 98: Please also check for `$end` here.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+                    'Unbounded closing of template: %s',
+                    $templateKey
+                ));
+            } else {
+                // substrings of form {{#foo, {{/foo {{#foo} or {{/foo} treated as a normal string
+                $buffer .= $char . $templateHandle . $templateKey;
+            }
+        }
+
+        if ($this->pos === $this->length && !empty($this->unclosed)) {
```
nilmerg, line 115: The `$this->unclosed` array isn't necessary. There is never more than one entry in it. It's also never required outside the current scope. Use `$for !== null` here instead and reset `$for` above instead of popping the value from the array.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+                break;
+            }
+
+            $this->pos++;
+            $next = $this->string[$this->pos] === '}';
+
+            if ($start && $next) {
+                array_push($this->unclosed, $templateKey);
+
+                if (in_array($templateKey, array_keys($this->templateArgs), true)) {
```
nilmerg, line 86: Please.. use `isset($this->templateArgs[$templateKey])` :pray:

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+                ));
+            } else {
+                // substrings of form {{#foo, {{/foo {{#foo} or {{/foo} treated as a normal string
+                $buffer .= $char . $templateHandle . $templateKey;
+            }
+        }
+
+        if ($this->pos === $this->length && !empty($this->unclosed)) {
+            throw new Exception(vsprintf(
+                'Unbounded opening of template: %s',
```
nilmerg, line 117: Use `Unbound` here. `Unbounded` isn't quite the correct term for this.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+                }
+            } elseif ($for === $templateKey) {
+                // close the template
+                $this->pos++;
+                array_pop($this->unclosed);
+                break;
+            } elseif ($end && $next) {
+                // throw exception for unbounded closing of templates
+                throw new Exception(sprintf(
+                    'Unbounded closing of template: %s',
```
nilmerg, line 106: `Unbound` as well, please.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+            } else {
+                $parentArgs[] = $val;
+            }
+        }
+
+        parent::__construct($format, $parentArgs);
+    }
+
+    /**
+     * @param null $for
```
nilmerg, line 52: Please add a method description above. And an empty line before and after. And not `null` but `string` as type hint.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+class TemplateString extends FormattedString
+{
+    /** @var array */
+    protected $templateArgs = [];
+
+    protected $pos = 0;
+
+    protected $string;
+
+    protected $length;
```
nilmerg, line 33: Please add doc blocks here.

src/TemplateString.php
```diff
@@ -0,0 +1,174 @@
...
+ * ```
+ * $info = new TemplateString(
+ *     'Follow the %s for more information on %s {{#mustache}}Mustache{{/mustache}}',
+ *     [
+ *         new Link('doc/html', 'HTML documentation'),
+ *         Html::tag('strong', 'HTML elements'),
+ *        [
+ *        'mustache' => Html::tag('span')
+ *        ]
+ *     ]
```
nilmerg, line 20: That example doesn't quite show the advantage of this template functionality. Use this instead:

```suggestion
* $info = TemplateString::create(
*      'Follow the {{#doc}}HTML documentation{{/doc}} for more information on {{#strong}}HTML elements{{/strong}}',
*      [
*          'doc'    => new Link(null, 'doc/html'),
*          'strong' => Html::tag('strong')
*      ]
```

nilmerg (changes requested):

nilmerg (commented):
I like what I see! :heart: 

Though, there's still a test failing.

nilmerg (approved):
It's like fine wine :wink: 

Thank you Ravi!

PR #52, opened 2021-07-26
Fix infinite loop when a wrapper wraps the element in an extra container

lippserd:


nilmerg (approved):
Didn't checked tests, but they run through and my short tests with DB-Web succeeded as well.

PR #55, opened 2021-10-01
LocalDateTimeElement: Fix func call to bool if `DateTime` is not valid

sukhwinder33445:
fix https://github.com/Icinga/icingadb-web/issues/386


src/FormElement/LocalDateTimeElement.php
```diff
@@ -26,6 +26,11 @@ public function setValue($value)
 
     public function getValueAttribute()
     {
+        // Workaround for https://github.com/Icinga/icingadb-web/issues/386
+        if (is_bool($this->value)) {
```
nilmerg, line 30: Why does it fix the issue if it's only a workaround? For what is the workaround? Why is the value a boolean?

nilmerg: Okay, then please remove the comment. It may only be a case for chrome, but that's irrelevant here.

The value is `false` because it couldn't be parsed I reckon. This parsing should already either be avoided or appropriately handled so that the value is never anything else than a `DateTime` object or `null`.

sukhwinder33445: The message already exists. Since the changed method returns null, the message appears automatically.



> the value is never anything else than a `DateTime` object or `null`.

Should I extend the class and add a method for it? 🤔

nilmerg: Form elements support dedicated validators, why not add one? Take a look at Icinga Web 2's `DateTimeValidator` and add that to https://github.com/Icinga/ipl-validator. Then add a new method `addDefaultValidators()` to `BaseFormElement` and call it inside `getValidators()`. `LocalDateTimeElement` then implements `addDefaultValidators()` and registers `ipl\Validator\DateTimeValidator`.

nilmerg (changes requested):
The issue is not fixed.

PR #57, opened 2021-11-22
Introduce PasswordElement class

yhabteab:


yhabteab:
@cla-bot check

nilmerg (changes requested):
Web 2's `ResourceConfigForm` doesn't include the stored password in the DOM. I think we should do the same but natively in the element itself. I believe this should work, since `ipl\Html\Form` can be (and in this case is) always being populated with the configuration:

* Don't use property `$value` of `ipl\Html\FormElement\Password` to render the `value` attribute, use a constant dummy password instead
  * Use an attribute callback for this
* Don't accept the dummy password as value for property `$value` (otherwise the real password is overridden and cannot be retrieved again by `getValue()`)

This should ensure that:

* A call to `setValue()` with the real password still stores it in property `$value` (Done by `FormElements::populate()`)
* A call to `getValue()` still returns the real password, if one got set by `setValue()` (Done by `FormElements::getValues()`)
* The form doesn't render the real password in the DOM, but shows that a password is set
* The form accepts a new password

src/Form.php
```diff
@@ -352,6 +370,11 @@ protected function onError()
         if (! $errors->isEmpty()) {
             $this->prependHtml($errors);
         }
+
+        // Since the validation state of this form can be true on the initial call, however,
+        // on subsequent calls by child methods can reset $the->isValid, so the
+        // elements need to be notified over again
+        $this->onValidate();
```
nilmerg, line 377: The only possible chance to reset `isValid` is to call `validate()` directly, which no-one does, to my knowledge. So why is this necessary?

nilmerg (changes requested):
There's more, but I'd like to tackle that in person.

src/FormElement/FormElements.php
```diff
@@ -188,7 +188,8 @@ public function registerElement(FormElement $element)
         $this->elements[$name] = $element;
 
         if (array_key_exists($name, $this->populatedValues)) {
-            $element->setValue($this->populatedValues[$name]);
+            $element->setValue(end($this->populatedValues[$name]));
+            $element->setValueCandidates($this->populatedValues[$name]);
```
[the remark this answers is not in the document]
nilmerg: What's the outcome here?

src/Form.php
yhabteab: See [here](https://github.com/Icinga/icingadb-web/blob/master/application/forms/ApiTransportForm.php#L97).

src/FormElement/FormElements.php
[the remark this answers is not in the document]
yhabteab: Eric has decided to introduce a custom interface and this will then be implemented by `BaseFormElement`.

src/FormElement/PasswordElement.php
```diff
@@ -0,0 +1,55 @@
...
+                }
+
+                return null;
+            }
+        );
+    }
+
+    public function onRegistered(Form $form)
+    {
+        $form->on(Form::ON_VALIDATE, function () use ($form) {
```
nilmerg, line 40: `$form` is passed as argument to the handler, use that instead.

nilmerg (changes requested):

src/FormElement/PasswordElement.php
yhabteab: Do you mean this `Form::ON_VALIDATE` => `$form::ON_VALIDATE` or something else? I did not understand the request change correctly.

nilmerg: `$form->on(Form::ON_VALIDATE, function () use ($form) {` vs `$form->on(Form::ON_VALIDATE, function ($form) {`

nilmerg (approved):

PR #60, opened 2022-01-24
Don't always render error messages even with correct select options

### LINKED ISSUE #47: "$value is not allowed here": Error message is displayed even with correct select options
## Describe the bug

During my tests for the [enhanced dashboards](https://github.com/Icinga/icingaweb2/pull/4323), this error message pops up upon the dashboards belonging a specific home once I switch to another home even though the options were reseted just basing on the selected home. Instead, a dashboard that was selected in the previous home shows up as not allowed value for the newer home before even loading the new options and it took me a very long time to find out the reason. And according to my current state, I can say that this is not due to the code on my end but rather to the ``setValue()`` method. To make it clearer I have included a few screenshots.

## Screenshots

First active dashboard home ``Transactions`` and selected dashboard ``Current Incidents``.

Bildschirmfoto 2021-05-25 um 11 48 35

Now switch to another home ``Banking``.

Bildschirmfoto 2021-05-25 um 11 51 43

As you can see, the ``setValue()`` method is either called before the new options are set, remembers the previous active value somehow, or is not necessary to print this error message at all.

yhabteab:
fixes #47

### DIFF
--- a/src/FormElement/SelectElement.php
+++ b/src/FormElement/SelectElement.php
@@ -38,10 +38,15 @@ class SelectElement extends BaseFormElement
     public function validate()
     {
         $value = $this->getValue();
-        if (! ($option = $this->getOption($value)) || $option->getAttributes()->has('disabled')) {
+        if (
+            $value !== null && (
+                ! ($option = $this->getOption($value))
+                || $option->getAttributes()->has('disabled')
+            )
+        ) {
             $this->valid = false;
             $this->addMessage("'$value' is not allowed here");
-        } elseif ($this->isRequired() && strlen($value) === 0) {
+        } elseif ($this->isRequired() && $value === null) {
             $this->valid = false;
         } else {
             parent::validate();
--- a/src/FormElement/FormElements.php
+++ b/src/FormElement/FormElements.php
@@ -338,6 +338,22 @@ trait FormElements
     }
 
     /**
+     * Clear populated value of the given element
+     *
+     * @param string $name
+     *
+     * @return $this
+     */
+    public function clearPopulatedValue($name)
+    {
+        if (! $this->hasBeenSubmitted() && isset($this->populatedValues[$name])) {
+            unset($this->populatedValues[$name]);
+        }
+
+        return $this;
+    }
+
+    /**
      * Add all elements from the given element collection
      *
      * @param Form|SubFormElement $form


nilmerg:
I don't think this is the correct solution. This only occurs because upon a autosubmit, the form is partially validated. There's nothing wrong with this. Also if an element is asked to validate, it should do so no matter the form's state.

If I'm not mistaken, the element will be valid in case a wrong value is set with `setValue()` without submitting the form. This is bad.

Please think of an alternative. (But don't validate in `setValue()` e9ad70cd9d6bc9dd8fd956efa052fd2ae7006805)

yhabteab:
The problem is, we don't have the ability to find out, that a select element can be dependent on another select element. And when the parent element is autosubmitted, previous value of the child element are also submitted. Seeing as you don't like the first solution, I have introduced a new `SelectElement` attribute called `dependsOn` which allows us to cache the `parent` selectelement name and get this filtered out using jquery.

nilmerg (changes requested):
This still has a rather big flaw: Consider a form with two auto-submitting elements, each with its own individual dependent element.

```
1st autosubmit
|_ dependent element

2nd autosubmit
|_ dependent element
```

Interacting with the second autosubmit causes the dependent element of the first one to be cleared.

This is due to the fact, that the dependency check doesn't involve the parent element at all. But I guess you asked yourself already that question: How should it?

I guess, not at all. Thinking about this, I only see one last option: Completely handing over the control to concrete forms by only providing the required tools.

A concrete form has all the knowledge. It knows when a default value has gone invalid and it knows which options are provided by a select element. By giving the concrete form now the ability to control which elements should be populated, it can solve the problem by itself.

```
if firstAutosubmitValue === 'optionA'
    firstOptions = ['a', 'b', 'c']
else
    firstOptions = ['d', 'e', 'f']
endif

if firstOptionPopulatedValue not in firstOptions
    form.clearPopulatedValue('firstOption')

form.addElement('firstAutosubmit', ...)
form.addElement('firstOption', ...)
```

`clearPopulatedValue()` should be a no-op in case the form has been submitted by the user. What's sent explicitly by the user must not be ignored.

src/Form.php
```diff
@@ -332,6 +332,22 @@ public function validatePartial()
...
     }
 
+    /**
+     * Clear populated value of the given element
+     *
+     * @param string $name
+     *
+     * @return $this
+     */
+    public function clearPopulatedValue($name)
```
nilmerg, line 342: `$this->populatedValues` originates from `FormElements`, please add this there instead.

src/FormElement/SelectElement.php
```diff
@@ -38,7 +38,12 @@ public function hasOption($value)
     public function validate()
     {
         $value = $this->getValue();
-        if (! ($option = $this->getOption($value)) || $option->getAttributes()->has('disabled')) {
+        if (
+            $value && (
```
nilmerg, line 42: Please check for `!== null` here, the empty string and a zero may both be valid option ids.

src/FormElement/SelectElement.php
```diff
@@ -38,7 +38,12 @@ public function hasOption($value)
...
+                ! ($option = $this->getOption($value))
+                || $option->getAttributes()->has('disabled')
+            )
+        ) {
             $this->valid = false;
             $this->addMessage("'$value' is not allowed here");
         } elseif ($this->isRequired() && strlen($value) === 0) {
```
nilmerg, line 49: `strlen($value) === 0` then has to change as well. If `$value` is NULL, it must not be passed to `strlen()`, but if the value is missing, it can actually only be NULL, so `$value === null` should be enough here.

nilmerg (changes requested):

src/Form.php
nilmerg: Think of it this way: `$form->hasBeenSubmitted()` is part of a public interface, `FormElements::$populatedValues` however is not. It's perfectly fine that a class which uses this trait implements a method to solve it's requirements. But it's bad design that it accesses a property from a trait. It's technically possible, but should be avoided IMHO.

nilmerg (approved):

PR #61, opened 2022-01-31
SubmitElement: Don't compare label to determine whether the submit el…

yhabteab:
…ement is pressed

Actually, when the submit element label is modified afterwards (after the submit element has been registered), it doesn't detect that the element is actually pressed (the form is submitted), as the button label is compared to the actual value of
the element to determine if it is pressed.

### DIFF
--- a/src/FormElement/SubmitElement.php
+++ b/src/FormElement/SubmitElement.php
@@ -40,7 +40,7 @@ class SubmitElement extends InputElement implements FormSubmitElement
 
     public function hasBeenPressed()
     {
-        return $this->getButtonLabel() === $this->getValue();
+        return (bool) $this->getValue();
     }
 
     public function isIgnored()


nilmerg:
Please show the code that is showing the problem you have without this change.

yhabteab:
> Please show the code that is showing the problem you have without this change.

Does it matter? It's not right the way it's implemented now and should be fixed regardless of whether anyone has a problem with it.  Anyway, the problem I have is within the enhanced-dashboards. But you can easily reproduce it by changing the label of a submit button from the controller.

Thomas-Gelf:
Hi @yhabteab,

please let me add two distinct concerns:

1. I wrote quite some forms wildly twisting submit elements based on this library, works fine. You do of course need to check whether a button has been pressed before you change the button. Alternatively remove the old element and register a new one, I sometimes shuffle buttons this way for small inline-forms with multiple navigation paths. When there are multiple submit elements in a form, I usually ask directly after adding the button, whether a specific submit element has been pressed. This allows for pleasant-to-read form code. Example: `if ($this->wantsToContinue()) { ... }`, with `wantsToContinue()` being a method which a) adds the input element and b) returns whether this element has been pressed. So I'm with @nilmerg, please show your specific use-case. There must be a better solution for you problem.
2. While point 1 is more a "What's the problem that needs to be solved"-kind of concern, there is a more severe problem. This pull request would break a completely valid use case: having multiple input elements with the same label in the very same form. That's perfectly valid as of the HTML standard.

Hope this helps!

Cheers,
Thomas

Thomas-Gelf:
Hint: updated the comment, it should have read *with the same **name*** (not label)

PR #65, opened 2022-07-13
FormElement: Add Multiselect element

sukhwinder33445:


src/FormElement/MultiSelectElement.php
```diff
@@ -0,0 +1,84 @@
...
+
+namespace ipl\Html\FormElement;
+
+use ipl\Html\Attributes;
+
+class MultiSelectElement extends SelectElement
+{
+    /**
+     * @var array Selected values
+     */
```
nilmerg, line 11: The compact version please.

src/FormElement/MultiSelectElement.php
```diff
@@ -0,0 +1,84 @@
...
+    protected $value = [];
+
+    public function __construct($name, $attributes = null)
+    {
+        parent::__construct($name, $attributes);
+        $this->getAttributes()->add('multiple', true);
```
nilmerg, line 17: If you initialize `BaseHtmlElement::$defaultAttributes` you don't need to override the constructor. (Preferred)

tests/FormElement/MultiSelectElementTest.php
```diff
@@ -0,0 +1,391 @@
...
+
+use ipl\Html\FormElement\MultiSelectElement;
+use ipl\Tests\Html\TestCase;
+
+class MultiSelectElementTest extends TestCase
+{
+    public function testFlatOptions()
+    {
+        $select = new MultiSelectElement('elname', [
+            'label'   => 'Customer',
```
nilmerg, line 13: What's that? Seems useless to me. Please remove it from all tests.

tests/FormElement/MultiSelectElementTest.php
```diff
@@ -0,0 +1,391 @@
...
+        $this->assertFalse($select->isValid());
+        $select->setValue(3);
+        $this->assertFalse($select->isValid());
+        $select->setValue(4);
+        $this->assertTrue($select->isValid());
+        $select->setValue('4');
+        $this->assertTrue($select->isValid());
+        $select->setValue('Down');
+        $this->assertTrue($select->isValid());
+        $select->setValue('Here');
```
nilmerg, line 97: Of course, a label can't be chosen, so this should test whether `sub` can be chosen or not.

tests/FormElement/MultiSelectElementTest.php
```diff
@@ -0,0 +1,391 @@
...
+                '1'  => 'The one',
+                '4'  => 'Four',
+                '5'  => 'Hi five'
+            ],
+        ]);
+
+        $this->assertFalse($select->isValid());
+    }
+
+    public function testOptionValidity()
```
nilmerg, line 73: Shouldn't this also test multiple selected options? There seems to be no other test for this at least.

nilmerg (changes requested):

src/FormElement/MultiSelectElement.php
```diff
@@ -0,0 +1,61 @@
...
+                break;
+            }
+        }
+
+        return $this;
+    }
+
+    protected function isSelectedOption($optionValue)
+    {
+        return in_array($optionValue, $this->getValue());
```
nilmerg, line 59: `in_array()` also supports loose and strict comparison, why not utilize it?

src/FormElement/MultiSelectElement.php
```diff
@@ -0,0 +1,61 @@
...
+        return $this;
+    }
+
+    public function validate()
+    {
+        BaseFormElement::validate();
+
+        if ($this->valid === false) {
+            return $this;
+        }
```
nilmerg, line 42: I think the element should have authority over what's valid in the first place. If it's a forbidden value, why validate it further? This may even pose a security risk depending on what's offered and how it's validated further.

src/FormElement/SelectElement.php
```diff
@@ -123,20 +123,29 @@ protected function makeOption($value, $label)
...
-                    : $this->getValue() === $optionValue;
+                return $this->isSelectedOption($option->getValue());
             });
             $this->options[$value] = $option;
 
             return $this->options[$value];
         }
     }
 
+    /** Whether the given option is a selected option
```
nilmerg, line 134: Doesn't this look ugly to you? To me it does.

nilmerg (changes requested):

src/FormElement/SelectElement.php
```diff
@@ -45,7 +48,7 @@ public function validate()
             )
         ) {
             $this->valid = false;
-            $this->addMessage("'$value' is not allowed here");
+            $this->addMessage(sprintf($this->translate('%s is not allowed here'), $value));
```
nilmerg, line 51: The single quotes around the value are removed now. I'd prefer to keep them. As everywhere else where you've changed this.

src/FormElement/SelectElement.php
```diff
@@ -67,9 +70,9 @@ public function disableOption($value)
         if ($option = $this->getOption($value)) {
             $option->getAttributes()->add('disabled', true);
         }
-        if ($this->getValue() == $value) {
+        if ($this->isSelectedOption($value)) {
```
nilmerg, line 73: Please insert an empty line here.

nilmerg (changes requested):

nilmerg (dismissed):

nilmerg (dismissed):

src/FormElement/MultiselectElement.php
```diff
@@ -0,0 +1,57 @@
...
+use ipl\Html\Attributes;
+
+class MultiselectElement extends SelectElement
+{
+    /** @var array Selected values */
+    protected $value = [];
+
+    protected $defaultAttributes = ['multiple' => true];
+
+    protected function registerValueCallback(Attributes $attributes)
```
lippserd, line 14: Please do it like this:

```php
    {
        // select elements don't have a value attribute
        return null;
    }

File: src/FormElement/MultiselectElement.php [hunk 12]
```diff
@@ -0,0 +1,57 @@
...
+namespace ipl\Html\FormElement;
+
+use ipl\Html\Attributes;
+
+class MultiselectElement extends SelectElement
+{
+    /** @var array Selected values */
+    protected $value = [];
+
+    protected $defaultAttributes = ['multiple' => true];
```
lippserd, line 12: This is an essential attribute and should therefore be added during `assemble()`.

src/FormElement/MultiselectElement.php
```diff
@@ -0,0 +1,57 @@
...
+            $option = $this->getOption($value);
+            if (! $option || $option->getAttributes()->has('disabled')) {
+                $this->valid = false;
+                $this->addMessage(sprintf($this->translate("'%s' is not allowed here"), $value));
+
+                return $this;
+            }
+        }
+
+        BaseFormElement::validate();
```
lippserd, line 48: Why not `parent`?

src/FormElement/SelectElement.php
```diff
@@ -45,13 +48,13 @@ public function validate()
             )
         ) {
             $this->valid = false;
-            $this->addMessage("'$value' is not allowed here");
-        } elseif ($this->isRequired() && $value === null) {
-            $this->valid = false;
-        } else {
-            parent::validate();
+            $this->addMessage(sprintf($this->translate("'%s' is not allowed here"), $value));
```
lippserd, line 51: Possible candidate for `InArrayValidator` or `DeferredInArrayValidator`.

lippserd (changes requested):
Please use [nowdoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc) for expected HTML.

src/FormElement/MultiselectElement.php
sukhwinder33445: Because the parent class checks its own value first, since I already have a loop above for this check, I skip calling the parent method.

src/FormElement/SelectElement.php
sukhwinder33445: I have already created a new [PR](https://github.com/Icinga/ipl-html/pull/83) for this.

lippserd: But I don't see its usage here.

src/FormElement/MultiselectElement.php
lippserd: Exactly. In addition we have duplicated code. This shows that the code flow is flawed, and it raises the question of why we even have a separate element for this. Instead, specifying the multiple attribute for the select element should just work.

lippserd (changes requested):

sukhwinder33445:
Add `multiple` attribute functionality to [SelectElement](https://github.com/Icinga/ipl-html/pull/88).

PR #66, opened 2022-07-15
FormElement: Add `FileElement`

sukhwinder33445:
## Blocked by

- [x] https://github.com/Icinga/ipl-html/pull/88 (trait MultipleAttribute required)

### DIFF
--- a/src/FormElement/FileElement.php
+++ b/src/FormElement/FileElement.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace ipl\Html\FormElement;
+
+use ipl\Html\Attributes;
+use Psr\Http\Message\UploadedFileInterface;
+use ipl\Html\Common\MultipleAttribute;
+
+class FileElement extends InputElement
+{
+    use MultipleAttribute;
+
+    protected $type = 'file';
+
+    /** @var UploadedFileInterface|UploadedFileInterface[] */
+    protected $value;
+
+    public function getValueAttribute()
+    {
+        // Value attributes of file inputs are set only client-side.
+        return null;
+    }
+
+    public function getNameAttribute()
+    {
+        $name = parent::getNameAttribute();
+
+        return $this->isMultiple() ? ($name . '[]') : $name;
+    }
+
+    public function hasValue()
+    {
+        if ($this->value === null) {
+            return false;
+        }
+
+        $file = $this->value;
+
+        if ($this->isMultiple()) {
+            return $file[0]->getError() !== UPLOAD_ERR_NO_FILE;
+        }
+
+        return $file->getError() !== UPLOAD_ERR_NO_FILE;
+    }
+
+    public function getValue()
+    {
+        return $this->hasValue() ? $this->value : null;
+    }
+
+    protected function registerAttributeCallbacks(Attributes $attributes)
+    {
+        parent::registerAttributeCallbacks($attributes);
+        $this->registerMultipleAttributeCallback($attributes);
+    }
+}
--- a/src/FormElement/FileElement.php
+++ b/src/FormElement/FileElement.php
@@ -3,6 +3,8 @@
 namespace ipl\Html\FormElement;
 
 use ipl\Html\Attributes;
+use ipl\Validator\FileValidator;
+use ipl\Validator\ValidatorChain;
 use Psr\Http\Message\UploadedFileInterface;
 use ipl\Html\Common\MultipleAttribute;
 
@@ -14,6 +16,9 @@ class FileElement extends InputElement
 
     /** @var UploadedFileInterface|UploadedFileInterface[] */
     protected $value;
+
+    /** @var int The default maximum file size */
+    protected static $defaultMaxFileSize;
 
     public function __construct($name, $attributes = null)
     {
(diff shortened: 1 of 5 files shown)


sukhwinder33445:
@cla-bot check

src/FormElement/FileElement.php
```diff
@@ -0,0 +1,51 @@
...
+        if ($this->getAttributes()->has('multiple')) {
+            return $this->getName() . '[]';
+        }
+
+        return parent::getNameAttribute();
+    }
+
+    public function onRegistered(Form $form)
+    {
+        $this->value = $form->getRequest()->getUploadedFiles();
```
lippserd, line 30: This is problematic for the following reasons:

1. it only works if `Form::handleRequest()` is called BEFORE the form is assembled. Any call to `Form::ensureAssembled()` before `Form::handleRequest()` will leave the value empty,
2. all uploaded files will be set to all file inputs, and
3. it does not work with nested form elements.
 
Read about the whole thing [here](https://www.php-fig.org/psr/psr-7/#16-uploaded-files).

I suggest doing the following in `Form::handleRequest()` before populating the `$params` to mitigate any issues mentioned above:

`$params = array_merge_recursive($params, $request->getUploadedFiles());`

src/FormElement/FileElement.php
```diff
@@ -0,0 +1,51 @@
...
+use ipl\Html\Form;
+
+class FileElement extends InputElement
+{
+    protected $type = 'file';
+
+    /** @var array of \GuzzleHttp\Psr7\UploadedFile objects */
+    protected $value = [];
+
+    public function setValue($value)
```
lippserd, line 14: `setValue()` is required for `populate()`. Do the following instead:

```php
    {
        // Value attributes of file inputs are set only client-side.
        return null;
    }
```

src/FormElement/FileElement.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->value = $form->getRequest()->getUploadedFiles();
+    }
+
+    public function hasValue()
+    {
+        if (empty($this->value)) {
+            return false;
+        }
+
+        $file = $this->value[$this->getName()];
```
lippserd, line 39: Consider the changes requested in `onRegistered()`.

src/FormElement/FileElement.php
```diff
@@ -0,0 +1,51 @@
...
+        if ($this->getAttributes()->has('multiple')) {
+            return $file[0]->getError() !== UPLOAD_ERR_NO_FILE;
+        }
+
+        return $file->getError() !== UPLOAD_ERR_NO_FILE;
+    }
+
+    public function getValue()
+    {
+        return $this->hasValue() ? $this->value[$this->getName()] : null;
```
lippserd, line 49: Consider the changes requested in `onRegistered()`.

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,40 @@
...
+namespace ipl\Tests\Html\FormElement;
+
+use ipl\Html\FormElement\FileElement;
+use ipl\Tests\Html\TestCase;
+
+class FileElementTest extends TestCase
+{
+    public function testFileElement()
+    {
+        $file = new FileElement('elname');
```
lippserd, line 12: Use `test_file` for the element name.

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace ipl\Tests\Html\FormElement;
+
+use ipl\Html\FormElement\FileElement;
+use ipl\Tests\Html\TestCase;
+
+class FileElementTest extends TestCase
+{
+    public function testFileElement()
```
lippserd, line 10: Rename to `testRendering()`.

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function testMultipleAttribute()
+    {
+        $file = new FileElement('elname', ['multiple' => true]);
+
+        $this->assertHtml('<input multiple name="elname[]" type="file"/>', $file);
+    }
+
+    public function testAcceptAttribute()
```
lippserd, line 24: You can remove this. This will not test anything special from the file input.

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,40 @@
...
+        $file = new FileElement('elname', ['accept' => 'image/jpeg, image/png']);
+
+        $this->assertHtml('<input name="elname" type="file" accept="image/jpeg, image/png"/>', $file);
+    }
+
+    public function testSetValueDoesNotEffectAnything()
+    {
+        $file = new FileElement('elname');
+
+        $this->assertHtml('<input name="elname" type="file"/>', $file);
```
lippserd, line 35: Already covered in `testRendering()`.

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function testAcceptAttribute()
+    {
+        $file = new FileElement('elname', ['accept' => 'image/jpeg, image/png']);
+
+        $this->assertHtml('<input name="elname" type="file" accept="image/jpeg, image/png"/>', $file);
+    }
+
+    public function testSetValueDoesNotEffectAnything()
```
lippserd, line 31: Rename to `testValueAttributeIsNotRendered()`

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace ipl\Tests\Html\FormElement;
+
+use ipl\Html\FormElement\FileElement;
+use ipl\Tests\Html\TestCase;
+
+class FileElementTest extends TestCase
```
lippserd, line 8: Add the following test first:

```
    public function testElementLoading(): void
    {
        $form = (new Form())
            ->addElement('file', 'test_file');

        $this->assertInstanceOf(FileElement::class, $form->getElement('test_file'));
    }
}

File: tests/FormElement/FileElementTest.php [hunk 11]
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function testFileElement()
+    {
+        $file = new FileElement('elname');
+
+        $this->assertHtml('<input name="elname" type="file"/>', $file);
+    }
+
+    public function testMultipleAttribute()
```
lippserd, line 17: Rename to `testMutipleAttributeAlsoChangesNameAttribute()` and assert that `getName()` still returns the original name.

lippserd (dismissed):
In addition to the changes requested, please test actual requests with `ServerRequestInterface:: withUploadedFiles()`.

src/FormElement/FileElement.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function getNameAttribute()
+    {
+        $name = parent::getNameAttribute();
+
+        return $this->isMultiple() ? ($name . '[]') : $name;
+    }
+
+    public function hasValue()
```
nilmerg, line 31: Here I wonder what happens if `$value` is null? Can it be null?

tests/FormElement/FileElementTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function testMutipleAttributeAlsoChangesNameAttribute()
+    {
+        $file = new FileElement('test_file', ['multiple' => true]);
+
+        $this->assertHtml('<input multiple name="test_file[]" type="file"/>', $file);
+        $this->assertSame($file->getName(), 'test_file');
+    }
+
+    public function testAcceptAttribute()
```
nilmerg, line 57: What purpose has this test? The `accept` attribute is not specifically processed by the element, so it's only subject to correct rendering. Though, for this, array syntax is missing. What about `true` and `false`? This is of course non-sense to assert here, it's a `FileElementTest` after all, not an `AttributeTest`. (Guess what, there's already such a test ;) )

Please remove this.

nilmerg (changes requested):
While looking at the `testAcceptAttribute` test I wonder if it makes sense to automatically add the `FileValidator`. Should be considered, IMHO.

src/FormElement/FileElement.php
sukhwinder33445: The `ServerRequest::normalizeFiles($_FILES)` call ensures that `ServerRequest::getUploadedFiles()` always returns `UploadedFileInterface|UploadedFileInterface[]`.

nilmerg: Okay, and what if no-one called `setValue()`?

sukhwinder33445: This can only happen if `Form::populate()` cannot find the element. Can this be the case at all?

nilmerg: You're talking about a chronological situation here. The fact that the value can technically be null should already be enough.

sukhwinder33445: Oh, got it. Thanks :)

PR #67, opened 2022-08-05
SelectElement:  Override existing options in `setOptions()`

sukhwinder33445:
`SelectElement::setOptions($opts)` call should not add $opts to existing options, but overwrite them completely.


lippserd (changes requested):
Please explain in the commit message (and description) what you are fixing here. Right now it reads like a feature.

lippserd (approved):

PR #68, opened 2022-08-15
FormElement: Add `RadioElement`

sukhwinder33445:
## Blocked by

* https://github.com/Icinga/ipl-validator/pull/12

### DIFF
--- a/src/FormElement/RadioElement.php
+++ b/src/FormElement/RadioElement.php
@@ -1,0 +1,177 @@
+<?php
+
+namespace ipl\Html\FormElement;
+
+use InvalidArgumentException;
+use ipl\Html\Attributes;
+use ipl\Html\HtmlDocument;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+use ipl\I18n\Translation;
+use ipl\Validator\DeferredInArrayValidator;
+use ipl\Validator\ValidatorChain;
+
+class RadioElement extends BaseFormElement
+{
+    use Translation;
+
+    /** @var string The element type */
+    protected $type = 'radio';
+
+    /** @var RadioOption[] Radio options */
+    protected $options = [];
+
+    /** @var array Disabled radio options */
+    protected $disabledOptions = [];
+
+    /**
+     * Set the options
+     *
+     * @param array $options
+     *
+     * @return $this
+     */
+    public function setOptions(array $options): self
+    {
+        $this->options = [];
+        foreach ($options as $value => $label) {
+            $option = (new RadioOption($value, $label))
+                ->setDisabled(
+                    in_array($value, $this->disabledOptions, ! is_int($value))
+                    || ($value === '' && in_array(null, $this->disabledOptions, true))
+                );
+
+            $this->options[$value] = $option;
+        }
+
+        $this->disabledOptions = [];
+
+        return $this;
+    }
+
+    /**
+     * Get the option with specified value
+     *
+     * @param string|int $value
+     *
+     * @return RadioOption
+     *
+     * @throws InvalidArgumentException If no option with the specified value exists
+     */
+    public function getOption($value): RadioOption
+    {
+        if (! isset($this->options[$value])) {
+            throw new InvalidArgumentException(sprintf('There is no such option "%s"', $value));
+        }
+
+        return $this->options[$value];
+    }
+
+    /**
+     * Set the specified options as disable
+     *
+     * @param array $disabledOptions
+     *
+     * @return $this
+     */
+    public function setDisabledOptions(array $disabledOptions): self
+    {
+        if (! empty($this->options)) {
+            foreach ($this->options as $value => $option) {
+                $option->setDisabled(
+                    in_array($value, $disabledOptions, ! is_int($value))
+                    || ($value === '' && in_array(null, $disabledOptions, true))
+                );
+            }
+
+            $this->disabledOptions = [];
+        } else {
+            $this->disabledOptions = $disabledOptions;
+        }
+
+        return $this;
+    }
+
+    public function renderUnwrapped()
+    {
+        // Parent::renderUnwrapped() requires $tag and the content should be empty. However, since we are wrapping
+        // each button in a label, the call to parent cannot work here and must be overridden.
+        return HtmlDocument::renderUnwrapped();
+    }
+
+    protected function assemble()
+    {
+        foreach ($this->options as $option) {
+            $radio = (new InputElement($this->getName()))
+                ->setType($this->type)
+                ->setValue($option->getValue());
+
+            // Only add the non-callback attributes to all options
+            foreach ($this->getAttributes() as $attribute) {
+                $radio->getAttributes()->addAttribute(clone $attribute);
+            }
+
+            $radio->getAttributes()
+                ->merge($option->getAttributes())
+                ->registerAttributeCallback(
+                    'checked',
+                    function () use ($option) {
+                        $optionValue = $option->getValue();
+
+                        return ! is_int($optionValue)
+                            ? $this->getValue() === $optionValue
+                            : $this->getValue() == $optionValue;
+                    }
+                )
+                ->registerAttributeCallback(
+                    'disabled',
+                    function () use ($option) {
+                        return $this->getAttributes()->get('disabled')->getValue() || $option->isDisabled();
+                    }
+                );
+
+            $label = new HtmlElement(
+                'label',
+                new Attributes(['class' => $option->getLabelCssClass()]),
+                $radio,
+                Text::create($option->getLabel())
+            );
+
+            $this->addHtml($label);
+        }
+    }
+
+    protected function addDefaultValidators(ValidatorChain $chain): void
+    {
+        $chain->add(new DeferredInArrayValidator(function (): array {
+            $possibleValues = [];
+
+            foreach ($this->options as $option) {
+                if ($option->isDisabled()) {
+                    continue;
+                }
+
+                $possibleValues[] = $option->getValue();
+            }
+
+            return $possibleValues;
+        }));
+    }
+
+    protected function registerAttributeCallbacks(Attributes $attributes)
+    {
+        parent::registerAttributeCallbacks($attributes);
+
+        $this->getAttributes()->registerAttributeCallback(
+            'options',
+            null,
+            [$this, 'setOptions']
+        );
+
+        $this->getAttributes()->registerAttributeCallback(
+            'disabledOptions',
+            null,
+            [$this, 'setDisabledOptions']
+        );
+    }
+}
--- a/src/FormElement/RadioOption.php
+++ b/src/FormElement/RadioOption.php
@@ -1,0 +1,148 @@
+<?php
+
+namespace ipl\Html\FormElement;
+
+use ipl\Html\Attributes;
+
+class RadioOption
+{
+    /** @var string The default label class */
+    public const LABEL_CLASS = 'radio-label';
+
+    /** @var string|int|null Value of the option */
+    protected $value;
+
+    /** @var string Label of the option */
+    protected $label;
+
+    /** @var mixed Css class of the option's label */
+    protected $labelCssClass = self::LABEL_CLASS;
+
+    /** @var bool Whether the radio option is disabled */
+    protected $disabled = false;
+
+    /** @var Attributes */
+    protected $attributes;
+
+    /**
+     * RadioOption constructor.
+     *
+     * @param string|int|null $value
+     * @param string $label
+     */
+    public function __construct($value, string $label)
+    {
+        $this->value = $value === '' ? null : $value;
+        $this->label = $label;
+    }
+
+    /**
+     * Set the label of the option
+     *
+     * @param string $label
+     *
+     * @return $this
+     */
+    public function setLabel(string $label): self
+    {
+        $this->label = $label;
+
+        return $this;
+    }
+
+    /**
+     * Get the label of the option
+     *
+     * @return string
+     */
+    public function getLabel(): string
+    {
+        return $this->label;
+    }
+
+    /**
+     * Get the value of the option
+     *
+     * @return string|int|null
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    /**
+     * Set css class to the option label
+     *
+     * @param string|string[] $labelCssClass
+     *
+     * @return $this
+     */
+    public function setLabelCssClass($labelCssClass): self
+    {
+        $this->labelCssClass = $labelCssClass;
+
+        return $this;
+    }
+
+    /**
+     * Get css class of the option label
+     *
+     * @return string|string[]
+     */
+    public function getLabelCssClass()
+    {
+        return $this->labelCssClass;
+    }
+
+    /**
+     * Set whether to disable the option
+     *
+     * @param bool $disabled
+     *
+     * @return $this
+     */
+    public function setDisabled(bool $disabled = true): self
+    {
+        $this->disabled = $disabled;
+
+        return $this;
+    }
+
+    /**
+     * Get whether the option is disabled
+     *
+     * @return bool
+     */
+    public function isDisabled(): bool
+    {
+        return $this->disabled;
+    }
+
+    /**
+     * Add the attributes
+     *
+     * @param Attributes $attributes
+     *
+     * @return $this
+     */
+    public function addAttributes(Attributes $attributes): self
+    {
+        $this->attributes = $attributes;
+
+        return $this;
+    }
+
+    /**
+     * Get the attributes
+     *
+     * @return Attributes
+     */
+    public function getAttributes(): Attributes
+    {
+        if ($this->attributes === null) {
+            $this->attributes = new Attributes();
+        }
+
+        return $this->attributes;
+    }
+}
--- a/tests/FormElement/RadioElementTest.php
+++ b/tests/FormElement/RadioElementTest.php
@@ -1,0 +1,454 @@
+<?php
+
+namespace ipl\Tests\Html\FormElement;
+
+use InvalidArgumentException;
+use ipl\Html\Attributes;
+use ipl\Html\Form;
+use ipl\Html\FormElement\RadioElement;
+use ipl\Html\FormElement\RadioOption;
+use ipl\I18n\NoopTranslator;
+use ipl\I18n\StaticTranslator;
+use ipl\Tests\Html\TestCase;
+
+class RadioElementTest extends TestCase
+{
+    public function testRendersElementCorrectly()
+    {
+        $radio = new RadioElement('test', [
+            'label'     => 'Test',
+            'options'   => [
+                'foo' => 'Foo',
+                'bar' => 'Bar',
+                'yes' => 'Yes'
+            ]
+        ]);
+
+        $html = <<<'HTML'
+<label class="radio-label"><input value="foo" name="test" type="radio">Foo</label>
+<label class="radio-label"><input value="bar" name="test" type="radio">Bar</label>
+<label class="radio-label"><input value="yes" name="test" type="radio">Yes</label>
+HTML;
+        $this->assertHtml($html, $radio);
+    }
+
+    public function testNumbersOfTypeIntOrStringAsOptionKeysAreHandledEqually()
+    {
+        $radio = new RadioElement('test', [
+            'label'     => 'Test',
+            'options'   => [
+                '1' => 'Foo',
+                2   => 'Bar',
+                3   => 'Yes'
+            ]
+        ]);
+
+        $radio->setValue(2);
+
+        $html = <<<'HTML'
+<label class="radio-label"><input value="1" name="test" type="radio">Foo</label>
+<label class="radio-label"><input value="2" name="test" type="radio" checked>Bar</label>
+<label class="radio-label"><input value="3" name="test" type="radio">Yes</label>
+HTML;
+        $this->assertHtml($html, $radio);
+
+        $radio->setValue('2');
+        $this->assertHtml($html, $radio);
+
+        $radio->setDisabledOptions([2]);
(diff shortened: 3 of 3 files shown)


src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,168 @@
...
+    public function setOptions(array $options): RadioElement
+    {
+        foreach ($options as $value => $label) {
+            $input = new InputElement($this->getName(), [
+                'name'  => $this->getName(),
+                'type'  => $this->type,
+                'value' => $value
+            ]);
+
+            $this->optionContent[$label] = $input;
```
lippserd, line 144: Don't index by label but value. That way you can just use `isset()` in other code places. Use label of the input element for the label 😉.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        return $this;
+    }
+
+    public function renderUnwrapped()
+    {
+        return $this->renderContent();
```
lippserd, line 152: Please justify this hack.

lippserd (changes requested):
In addition to the requested changes please

* use heredoc for the expected HTML and
* replace validation with something like an `InArray` validator provided in `addDefaultValidators()`.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,168 @@
...
+{
+    use Translation;
+
+    protected $type = 'radio';
+
+    protected $optionContent;
+
+    public function __construct($name, $attributes = null)
+    {
+        $this->setName($name);
```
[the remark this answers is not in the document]
sukhwinder33445: The `setOptions` setter callback requires $name before the parent constructor is even called.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,168 @@
...
+
+        // ZF1 compatibility:
+        $this->getAttributes()->registerAttributeCallback(
+            'multiOptions',
+            null,
+            [$this, 'setOptions']
+        );
+
+        $this->getAttributes()->registerAttributeCallback(
+            'disabled',
```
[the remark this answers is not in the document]
sukhwinder33445: Since each radio has its own value, label and disable attribute, should `disabled => true` disable all? 🤔

src/FormElement/RadioElement.php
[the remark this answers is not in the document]
sukhwinder33445: Fixed. The parent constructor should call `setName()` before attributes are added.

src/FormElement/RadioElement.php
sukhwinder33445: `Parent::renderUnwrapped()` requires `$tag` and the content should be empty. However, since we are creating each button in a label, the call to parent cannot work here and must be overridden.

sukhwinder33445:
> * replace validation with something like an `InArray` validator provided in `addDefaultValidators()`.

This will not help much, since we also need to check for the `disabled` attribute. And all this should be done before parent::validate() is called. 🤔

src/FormElement/RadioElement.php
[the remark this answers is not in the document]
lippserd: > [...], should `disabled => true` disable all?

Yes.

src/FormElement/RadioElement.php
lippserd: As comment in the code please 😉.

lippserd:
> > * replace validation with something like an `InArray` validator provided in `addDefaultValidators()`.
> 
> This will not help much, since we also need to check for the `disabled` attribute. And all this should be done before parent::validate() is called. 🤔

The value should be validated against a list of possible values. Possible values are all options that are not disabled.

lippserd:
Please don't resolve conversations just for fun. I don't see anything resolved in the code.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,168 @@
...
+    {
+        $this->setName($name);
+
+        $this->getAttributes()->registerAttributeCallback(
+            'options',
+            null,
+            [$this, 'setOptions']
+        );
+
+        // ZF1 compatibility:
```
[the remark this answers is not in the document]
sukhwinder33445: Removed.

lippserd (dismissed):
It should be `InArrayValidator` and not `InArray`, right?

I think it is best to have an option to defer which values to validate against in the `InArrayValidator`, by passing a callback to retrieve the possible values.

lippserd:
> It should be `InArrayValidator` and not `InArray`, right?
> 
> I think it is best to have an option to defer which values to validate against in the `InArrayValidator`, by passing a callback to retrieve the possible values.

Maybe `DeferredInArrayValidator`.

TAINCER:
This currently doesn't support adding custom `class` Attribute values to the Elements. Thus use cases where rendering different content based on the selection is not possible because it would need the class `autosubmit` attached to each Radio Input.

Would this be something to include?

yhabteab:
The element attributes also depend on the order in which they are added. E.g.:

This doesn't preselect the `each` option as the options are rendered after `setValue()`, I guess!
```php
$runsEach = $this->createElement('radio', 'runsOn', [
      'required' => true,
      'value'    => $this->getPopulatedValue('runsOn', 'each'),
      'options'  => ['each' => t('Each')],
]);
```

But this does preselect the option:

```php
$runsEach = $this->createElement('radio', 'runsOn', [
      'required' => true,
      'options'  => ['each' => t('Each')],
      'value'    => $this->getPopulatedValue('runsOn', 'each'),
]);
```

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        return $this->options[$value];
+    }
+
+    public function renderUnwrapped()
+    {
+        // Parent::renderUnwrapped() requires $tag and the content should be empty. However, since we are wrapping
+        // each button in a label, the call to parent cannot work here and must be overridden.
+        return $this->renderContent();
```
nilmerg, line 61: Please use `HtmlDocument::renderUnwrapped()`. The implementation of `renderContent` may change.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    protected function assemble()
+    {
+        foreach ($this->options as $option) {
+            $radio = (new InputElement($this->getName()))
+                ->setType($this->getType())
+                ->setValue($option->getValue())
+                ->setAttributes(clone $this->getAttributes());
```
nilmerg, line 70: I'm not sure this is expected by the caller. Not every attribute is applicable to all options at the same time. Just think of the `id` attribute.

Important attributes can be adjusted by interacting with the options directly. (required, disabled, etc.) For what else should this be necessary?

Also, cloning attributes doesn't work correctly currently.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,135 @@
...
+                ->registerAttributeCallback('disabled',
+                    function () use ($option) {
+                        return $option->isDisabled();
+                    }
+                )
+                ->registerAttributeCallback('class',
+                    function () use ($option) {
+                        // RadioOption::setCssClass() accepts a mixed param, so it can be an array, a string
+                        // or an attribute instance.
+                        // But the attribute callback accepts limited return types.
```
nilmerg, line 92: It cannot return arrays. That's the only reason so please mention it.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,135 @@
...
+                        return Attributes::create(['class', $option->getCssClass()])->get('class');
+                    }
+                );
+
+            $labelElm = new HtmlElement(
```
nilmerg, line 97: Please don't use abbreviations in variable names. `$label` or `$labelElement`.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,135 @@
...
+                'label',
+                Attributes::create(['class' => 'radio-label']),
+                $radio,
+                HtmlString::create($option->getLabel())
```
nilmerg, line 101: `Text::create()` please. `HtmlString::create()` does not escape.

src/FormElement/RadioOption.php
```diff
@@ -0,0 +1,155 @@
+<?php
+
+namespace ipl\Html\FormElement;
+
+use ipl\Html\Attribute;
+
+class RadioOption
```
nilmerg, line 7: Setters before getters please. Your arrangement currently isn't consistent as well.

tests/FormElement/RadioElementTest.php
```diff
@@ -0,0 +1,150 @@
...
+<label class="radio-label"><input value="bar" name="test" type="radio" checked>Bar</label>
+<label class="radio-label"><input value="yes" name="test" type="radio">Yes</label>
+<label class="radio-label"><input value="no" name="test" type="radio" disabled>No</label>
+HTML;
+        $this->assertHtml($html, $radio);
+    }
+
+    public function testRadioNotValidIfCheckedValueIsInvalid()
+    {
+       $this->markTestSkipped('Requires https://github.com/Icinga/ipl-validator/pull/8');
```
nilmerg, line 108: Don't forget to remove this.

tests/FormElement/RadioElementTest.php
```diff
@@ -0,0 +1,150 @@
...
+        $radio->setValue('no');
+        $this->assertFalse($radio->isValid());
+
+        $radio->setValue('tom');
+        $this->assertFalse($radio->isValid());
+    }
+
+    public function testRadioNotValidIfCheckedValueIsDisabled()
+    {
+        $this->markTestSkipped('Requires https://github.com/Icinga/ipl-validator/pull/8');
```
nilmerg, line 132: Don't forget to remove this.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @param string|int $value
+     *
+     * @return RadioOption
+     *
+     * @throws InvalidArgumentException If no option with the specified value exists
+     */
+    public function getOption($value): RadioOption
+    {
+        foreach ($this->options as $option) {
```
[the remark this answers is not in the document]
nilmerg: Ah, just saw your wip commit. Yes, I don't think a setter for value is required. There's `setOptions` to change options.

nilmerg (changes requested):
Since this has multiple values as well, please add a test like mentioned [here](https://github.com/Icinga/ipl-html/pull/88#discussion_r1032140457)

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,162 @@
...
+                ->registerAttributeCallback('disabled',
+                    function () use ($option) {
+                        return $this->getAttributes()->get('disabled')->getValue()
+                            || in_array($option->getValue(), $this->disabledOptions, true)
+                            || $option->isDisabled();
+                    }
+                );
+
+            $labelClass = (new Attributes(['class' => RadioOption::LABEL_CLASS]))
+                ->merge(new Attributes(['class' => $option->getlabelCssClass()]));
```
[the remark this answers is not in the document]
sukhwinder33445: I had this before. But add does not support `$value` instance of `Attribute`.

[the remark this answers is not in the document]
nilmerg: I see. But then I'd question why should this be possible:

`$form->getElement('radio')->getOption('foo')->setLabelCssClass(new Attribute('required', true))`

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,175 @@
...
+                    }
+                )
+                ->registerAttributeCallback('disabled',
+                    function () use ($option) {
+                        return $this->getAttributes()->get('disabled')->getValue() || $option->isDisabled();
+                    }
+                );
+
+            $labelAttributes = (new Attributes(['class' => RadioOption::LABEL_CLASS]))
+                ->add('class', $option->getlabelCssClass());
```
nilmerg, line 129: There's still this typo.

src/FormElement/RadioOption.php
```diff
@@ -0,0 +1,148 @@
...
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    /**
+     * Add css class to the option label
+     *
+     * @param string|array|Attribute $labelCssClass
```
nilmerg, line 76: The types aren't in line with the getter.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,175 @@
...
+     * @param array $options
+     *
+     * @return $this
+     */
+    public function setOptions(array $options): self
+    {
+        $this->options = [];
+        foreach ($options as $value => $label) {
+            $option = (new RadioOption($value, $label))
+                ->setDisabled(in_array($value, $this->disabledOptions, true));
```
nilmerg, line 39: I'm not sure `in_array` is suitable here. Remember, `''` and `null` should be equally handled.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,179 @@
...
+     */
+    public function setOptions(array $options): self
+    {
+        $this->options = [];
+        foreach ($options as $value => $label) {
+            $option = (new RadioOption($value, $label))
+                ->setDisabled(
+                    in_array($value, $this->disabledOptions, ! is_int($value))
+                    || (($value === '' || $value === null)
+                        && in_array(null, $this->disabledOptions, ! is_int($value)))
```
nilmerg, line 42: `$value` can't be `null`, as I showed you yesterday on the php-cli

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @param array $disabledOptions
+     *
+     * @return $this
+     */
+    public function setDisabledOptions(array $disabledOptions): self
+    {
+        if (($key = array_search('', $disabledOptions)) !== false) {
+            $disabledOptions[$key] = null;
+        }
```
nilmerg, line 82: What's this? Tests succeed without it.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,179 @@
...
+
+        if (! empty($this->options)) {
+            foreach ($this->options as $value => $option) {
+                $option->setDisabled(
+                    in_array($value, $disabledOptions, ! is_int($value))
+                    || (($value === '' || $value === null) && in_array(null, $disabledOptions, ! is_int($value)))
```
nilmerg, line 88: Same here, `$value` can't be `null`

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,179 @@
...
+            // Only add the non-callback attributes to all options
+            foreach ($this->getAttributes() as $attribute) {
+                $radio->getAttributes()->addAttribute(clone $attribute);
+            }
+
+            $radio->getAttributes()
+                ->merge($option->getAttributes())
+                ->registerAttributeCallback('checked',
+                    function () use ($option) {
+                        return $this->getValue() === ($option->getValue() === '' ? null : $option->getValue());
```
nilmerg, line 123: `getOption` shouldn't have the empty string as value, the element doesn't as well after all.

nilmerg (changes requested):
Missing tests:

* `testSetOptionsResetsOptions()`
* `testOrderOfOptionsAndDisabledOptionsDoesNotMatter()`
* `testGetOptionReturnsPreviouslySetOption()`
* ```php
    public function testNullAndTheEmptyStringAreAlsoEquallyHandledWhileDisablingOptions()
    {
        $radio = new RadioElement('radio');
        $radio->setOptions([null => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions([null]);

        $this->assertTrue($radio->getOption(null)->isDisabled());

        $radio = new RadioElement('radio');
        $radio->setOptions(['' => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions(['']);

        $this->assertTrue($radio->getOption('')->isDisabled());

        $radio = new RadioElement('radio');
        $radio->setOptions([null => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions(['']);

        $this->assertTrue($radio->getOption(null)->isDisabled());
        $radio = new RadioElement('radio');
        $radio->setOptions(['' => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions([null]);

        $this->assertTrue($radio->getOption('')->isDisabled());
    }
  ```
* ```php

    public function testGetOptionGetValueAndElementGetValueHandleNullAndTheEmptyStringEqually()
    {
        $radio = new RadioElement('radio');
        $radio->setOptions(['' => 'Foo']);
        $radio->setValue('');

        $this->assertNull($radio->getValue());
        $this->assertNull($radio->getOption('')->getValue());

        $radio = new RadioElement('radio');
        $radio->setOptions([null => 'Foo']);

        $this->assertNull($radio->getValue());
        $this->assertNull($radio->getOption(null)->getValue());
    }
  ```

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @return $this
+     */
+    public function setDisabledOptions(array $disabledOptions): self
+    {
+        if (! empty($this->options)) {
+            foreach ($this->options as $value => $option) {
+                $option->setDisabled(
+                    in_array($value, $disabledOptions, ! is_int($value))
+                    || ($value === '' && in_array(null, $disabledOptions))
```
nilmerg, line 83: Again, must be a strict comparison.

src/FormElement/RadioOption.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public function getLabel(): string
+    {
+        return $this->label;
+    }
+
+    /**
+     * Get the value of the option
+     *
+     * @return int|string
```
nilmerg, line 68: and... `null`.

tests/FormElement/RadioElementTest.php
```diff
@@ -0,0 +1,410 @@
...
+        $radio = new RadioElement('radio');
+        $radio->setOptions(['foo' => 'Foo', 'bar' => 'Bar']);
+
+        $this->assertInstanceOf(RadioOption::class, $radio->getOption('foo'));
+        $this->assertInstanceOf(RadioOption::class, $radio->getOption('bar'));
+
+        $radio->setOptions(['car' => 'Car', 'train' => 'Train']);
+
+        $this->expectExceptionMessage('There is no such option "foo"');
+        $radio->getOption('foo');
```
nilmerg, line 311: I don't think the test continues running after this.

nilmerg (changes requested):

tests/FormElement/RadioElementTest.php
sukhwinder33445: Yes, you are right. Moved it to the end of the test.

src/FormElement/RadioOption.php
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Add css class to the option label
+     *
+     * @param string|string[] $labelCssClass
+     *
+     *  @return $this
+     */
+    public function addLabelCssClass($labelCssClass): self
```
nilmerg, line 80: Rename this to `setLabelCssClass` please. The passed value should replace the default.

Anyone who still wants the default class, can do this: `$option->setLabelCssClass([RadioOption::LABEL_CLASS, 'new'])`

We don't provide a default style for `radio-label`, so there's nothing one could want to *keep*. And my suggestion recently for Yonas' problem would have the default made worse, because he would need to revert most of the CSS rules defined for the default class in order to hide the option. Even worse it could lead to this: `.radio-label:not(.hidden)`

So, being able to remove the default should be possible. Keeping the default also, which is the case by having the constant public.

nilmerg (changes requested):

src/FormElement/RadioOption.php
```diff
@@ -0,0 +1,147 @@
...
+    {
+        return $this->value;
+    }
+
+    /**
+     * Add css class to the option label
+     *
+     * @param string|string[] $labelCssClass
+     *
+     *  @return $this
```
nilmerg, line 77: ~~Add~~Set

The `@return` has a superfluous space at the beginning.

src/FormElement/RadioOption.php
```diff
@@ -0,0 +1,147 @@
...
+    public const LABEL_CLASS = 'radio-label';
+
+    /** @var string|int|null Value of the option */
+    protected $value;
+
+    /** @var string Label of the option */
+    protected $label;
+
+    /** @var mixed Css class of the option's label */
+    protected $labelCssClass = self::LABEL_CLASS;
```
nilmerg, line 18: Maybe use `static` instead. To be able to indeed subclass this and just override the constant.

src/FormElement/RadioElement.php
```diff
@@ -0,0 +1,173 @@
...
+                ->registerAttributeCallback(
+                    'disabled',
+                    function () use ($option) {
+                        return $this->getAttributes()->get('disabled')->getValue() || $option->isDisabled();
+                    }
+                );
+
+            $label = new HtmlElement(
+                'label',
+                (new Attributes(['class' => $option->getLabelCssClass()])),
```
nilmerg, line 131: You don't need the parentheses around this.

src/FormElement/RadioOption.php
sukhwinder33445: It does not work. I get the hint mesaage `'static::' is not allowed in compile time constants`.

nilmerg (approved):

yhabteab:
Please add an option to render the input fields without wrapping them into the `radio-label`.

sukhwinder33445:
> Please add an option to render the input fields without wrapping them into the `radio-label`.

Can you explain the use case please?

yhabteab:
I have a very special use case though that would stretch the scope to explain here but in short, I don't want the inputs to be wrapped in a label that I can neither access it with `getWrapper()` nor override it.

PR #69, opened 2022-08-18
Checkbox: Allow overriding `checked` and `unchecked` values via attributes

sukhwinder33445:


lippserd (changes requested):
Since overriding those values is already possible via setters the following commit message is more accurate:

> Checkbox: Allow overriding `checked` and `unchecked` values via attributes

lippserd (approved):

PR #70, opened 2022-08-24
SelectOption: Add method `getLabel()` and change method `setLabel()`

sukhwinder33445:
This PR allows to easily get the label of the selected option. `getContect()` returns an array of `ValidHtml[]` and makes it harder to get the label as string.

# Usage:
`$form->getElement('user')->getOption($form->getValue('user'))->getLabel()`

### DIFF
--- a/src/FormElement/SelectOption.php
+++ b/src/FormElement/SelectOption.php
@@ -8,38 +8,62 @@ class SelectOption extends BaseHtmlElement
 {
     protected $tag = 'option';
 
-    /** @var mixed */
+    /** @var string|int|null Value of the option */
     protected $value;
+
+    /** @var string Label of the option */
+    protected $label;
 
     /**
      * SelectOption constructor.
-     * @param string|null $value
-     * @param string|null $label
+     *
+     * @param string|int|null $value
+     * @param string $label
      */
-    public function __construct($value = null, $label = null)
+    public function __construct($value, string $label)
     {
         $this->value = $value;
-        $this->add($label);
+        $this->label = $label;
 
         $this->getAttributes()->registerAttributeCallback('value', [$this, 'getValue']);
     }
 
     /**
-     * @param $label
+     * Set the label of the option
+     *
+     * @param string $label
+     *
      * @return $this
      */
-    public function setLabel($label)
+    public function setLabel(string $label): self
     {
-        $this->setContent($label);
+        $this->label = $label;
 
         return $this;
     }
 
     /**
+     * Get the label of the option
+     *
      * @return string
+     */
+    public function getLabel(): string
+    {
+        return $this->label;
+    }
+
+    /**
+     * Get the value of the option
+     *
+     * @return string|int|null
      */
     public function getValue()
     {
         return $this->value;
+    }
+
+    protected function assemble()
+    {
+        $this->setContent($this->getLabel());
     }
 }
--- a/tests/FormElement/SelectElementTest.php
+++ b/tests/FormElement/SelectElementTest.php
@@ -257,11 +257,4 @@ class SelectElementTest extends TestCase
             $select
         );
     }
-
-    public function testLabelCanBeChanged()
-    {
-        $option = new SelectOption('value', 'Original label');
-        $option->setLabel('New label');
-        $this->assertHtml('<option value="value">New label</option>', $option);
-    }
 }
--- a/tests/FormElement/SelectOptionTest.php
+++ b/tests/FormElement/SelectOptionTest.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace ipl\Tests\Html\FormElement;
+
+use ipl\Html\FormElement\SelectElement;
+use ipl\Html\FormElement\SelectOption;
+use ipl\Tests\Html\TestCase;
+
+class SelectOptionTest extends TestCase
+{
+    public function testRendering()
+    {
+        $option = new SelectOption('test', 'Original label');
+
+        $this->assertHtml('<option value="test">Original label</option>', $option);
+    }
+
+    public function testRenderingAfterSetLabel()
+    {
+        $option = new SelectOption('test', 'Original label');
+        $option->setLabel('New label');
+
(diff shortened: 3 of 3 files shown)


nilmerg:
Please rebase here, there's a new test covering this change.

lippserd (commented):
What is the purpose of this PR, i.e. why do you have to change the label after construction? Also, there's `setContent()` and `getContent()` which could just be used on the element.

sukhwinder33445:
I noticed that label can be null in the constructor. It makes no sense to have an option without label. Maybe this should be fixed in a separate PR.

lippserd:
> What is the purpose of this PR, i.e. why do you have to change the label after construction? Also, there's `setContent()` and `getContent()` which could just be used on the element.

So you have updated the description. I suppose now it's just a matter of `getLabel()`. What do you want do with the label?

lippserd:
> I noticed that label can be null in the constructor. It makes no sense to have an option without label. Maybe this should be fixed in a separate PR.

We'll talk about that later. option can be used in several places and here and there it applies to have no value. For a select option, of course, this makes no sense. However, value could be empty, which then defaults to the content. I will create an issue for all this later. First, I want to understand what you want to do with the label.

lippserd:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

lippserd:
@sukhwinder33445 This can be closed, right?

lippserd (changes requested):
In addition to the requested changes, `setLabel()` must NOT be removed, see [here for a usage example](https://github.com/Icinga/icingaweb2-module-vspheredb/blob/master/library/Vspheredb/Web/Form/InfluxDbConnectionForm.php#L178).

 I would like to see the following tests:

* `testRendering`
* `testRenderingAfterSetLabel`
* `testGetLabel`
* `testGetLabelAfterSetLabel`

src/FormElement/SelectOption.php
```diff
@@ -19,27 +22,45 @@ class SelectOption extends BaseHtmlElement
...
+        $this->label = $label;
 
         $this->getAttributes()->registerAttributeCallback('value', [$this, 'getValue']);
     }
 
     /**
-     * @param $label
+     * Set the label of the option
+     *
+     * @param string
```
nilmerg, line 33: `@param` requires also the name of the param

nilmerg (changes requested):

nilmerg (approved):

src/FormElement/SelectOption.php
```diff
@@ -11,6 +11,9 @@ class SelectOption extends BaseHtmlElement
     /** @var mixed */
     protected $value;
 
+    /** @var string Label of the option */
```
lippserd, line 14: `string|null`

src/FormElement/SelectOption.php
```diff
@@ -19,27 +22,45 @@ class SelectOption extends BaseHtmlElement
...
 
     /**
-     * @param $label
+     * Set the label of the option
+     *
+     * @param string $label
+     *
      * @return $this
      */
     public function setLabel($label)
```
lippserd, line 37: `string $label` or `string $label = null` if we want to support resetting the label.

lippserd: Add return type.

lippserd (changes requested):

nilmerg:
As we're changing this so fundamentally, may I ask what the purpose of a `SelectOption` without a value is? Once constructed, it's not possible to change it, so if not passed to the constructor, the value is permanently not set.
I'd also question that the label may not be set. What's the purpose of a `SelectOption` without a label?

lippserd:
> As we're changing this so fundamentally, may I ask what the purpose of a `SelectOption` without a value is? Once constructed, it's not possible to change it, so if not passed to the constructor, the value is permanently not set. I'd also question that the label may not be set. What's the purpose of a `SelectOption` without a label?

`null` is already used as value: https://github.com/Icinga/icingaweb2-module-vspheredb/blob/master/library/Vspheredb/Web/Form/InfluxDbConnectionForm.php#L60

Not having a label is questionable though.

src/FormElement/SelectOption.php
```diff
@@ -8,38 +8,61 @@ class SelectOption extends BaseHtmlElement
...
+    /** @var string|null Label of the option */
+    protected $label;
+
     /**
      * SelectOption constructor.
      * @param string|null $value
      * @param string|null $label
      */
-    public function __construct($value = null, $label = null)
+    public function __construct(string $value = null, string $label = null)
```
nilmerg, line 22: Then I'd recommend this: (and the consequences for the other parts that follow)

```suggestion
    public function __construct(?string $value, string $label)
```

src/FormElement/SelectOption.php
```diff
@@ -8,38 +8,62 @@ class SelectOption extends BaseHtmlElement
 {
     protected $tag = 'option';
 
-    /** @var mixed */
+    /** @var string|null Value of the option */
```
nilmerg, line 11: Please use `?string` here as well.

src/FormElement/SelectOption.php
```diff
@@ -8,38 +8,62 @@ class SelectOption extends BaseHtmlElement
...
     protected $value;
 
+    /** @var string Label of the option */
+    protected $label;
+
     /**
      * SelectOption constructor.
+     *
      * @param string|null $value
```
nilmerg, line 20: Please use `?string`

src/FormElement/SelectOption.php
```diff
@@ -8,38 +8,62 @@ class SelectOption extends BaseHtmlElement
...
      * @return string
      */
-    public function getValue()
+    public function getLabel(): string
+    {
+        return $this->label;
+    }
+
+    /**
+     * Get the label of the option
```
nilmerg, line 56: The label?

nilmerg (changes requested):

nilmerg (approved):

PR #71, opened 2022-08-25
Add Collection for dynamic Forms

TAINCER:
This aims to add support for dynamic Form Fields. The API is still WIP, up to date examples can be found in `CollectionTest.php`.

Simple Example:
```php
        $collection = new Collection('testCollection');

        $collection->setAddTrigger('select', 'add_element', [
            'required' => false,
            'label'    => 'Add Trigger',
            'options'  => [null => 'Please choose', 'first' => 'First Option'],
            'class'    => 'autosubmit'
        ]);

        $collection->on(Collection::ON_LOAD, function ($group, $addElement, $_removeElement) {
            $group->addElement($addElement);

            $group->addElement('input', 'test_input');
        });
```


lippserd (changes requested):
* Please always use single quotes for strings that don't require interpolation. This helps to distinguish strings that actually require it.
* Use heredoc/nowdoc for expected HTML.
* Always add a newline at the end of files. PHPStorm has a setting for this. Also, PHPStorm can automatically remove spaces at the end of lines.
* I would go with the following API:
    * `setTriggerElement()` instead of `setAddTrigger()`.
    * `setRemoveElement()` instead of `setRemoveTrigger()`.
    * `onTrigger()` instead of `on(Collection::ON_LOAD, ...)`.
* Remove events as they become obsolete with the above changes.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+
+        if (! $hasEmptyTriggers) {
+            $this->value[] = [];
+            $this->emit(self::ON_ADD);
+        }
+
+        foreach ($this->value ?? [[]] as $key => $items) {
+            if (isset($items['__remove_entry'])) {
```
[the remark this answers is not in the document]
TAINCER: Yes, but it should use the key from the groups `remoteElement`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+
+    protected function assemble()
+    {
+        $this->setAttribute('class', 'collection');
+
+        $this->decorator = $this->setDefaultElementDecorator(new IcingaFormDecorator());
+
+        $this->addHtml(new HtmlElement('legend', null, new Text($this->getLabel())));
+
+        $this->value = $this->getPopulatedValues() ?? [[]];
```
lippserd, line 57: Don't use the protected `$value` property here, as fieldsets/collections shouldn't set this. Just use a local variable called `$values`.  If the populated values are empty, `getPopulatedValues()` returns an empty array. This should be sufficient, see below for an explanation.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+
+        $isEmpty = false;
```
lippserd, line 59: Please use `$valid = true`. The explanation follows.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+        foreach ($this->value as $key => $items) {
+            if (isset($items[0][$this->removeTrigger['name']]) && count($this->value) > 1) {
+                continue;
+            }
+
+            $group = $this->addGroup($key);
+
+            if (empty($group->getPopulatedValue($this->getName()))) {
+                $isEmpty = true;
```
lippserd, line 68: Set `$valid = false`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+
+        if (! $isEmpty) {
```
lippserd, line 72: Use `if ($valid) ...`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+
+        $this->decorator = $this->setDefaultElementDecorator(new IcingaFormDecorator());
+
+        $this->addHtml(new HtmlElement('legend', null, new Text($this->getLabel())));
+
+        $this->value = $this->getPopulatedValues() ?? [[]];
+
+        $isEmpty = false;
+        foreach ($this->value as $key => $items) {
+            if (isset($items[0][$this->removeTrigger['name']]) && count($this->value) > 1) {
```
lippserd, line 61: Consider the changes here with `$removeElement` and its new default value of `null`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+        $this->setAttribute('class', 'collection');
+
+        $this->decorator = $this->setDefaultElementDecorator(new IcingaFormDecorator());
+
+        $this->addHtml(new HtmlElement('legend', null, new Text($this->getLabel())));
+
+        $this->value = $this->getPopulatedValues() ?? [[]];
+
+        $isEmpty = false;
+        foreach ($this->value as $key => $items) {
```
lippserd, line 60: Use `foreach ($values ...)`.

Justification for the above changes:

You use `[[]]` as empty values so that the loop runs at least once to add an element. Now, with `$valid = true` the loop doesn't have to run once and `[]` from `getPopulatedValues()` is sufficient.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+
+        if (isset($this->removeTrigger['type']) && isset($this->removeTrigger['name'])) {
+            $removeElement = $this->createElement(
+                $this->removeTrigger['type'],
+                $this->removeTrigger['name'],
+                $this->removeTrigger['attributes']
+            );
+        }
+
+        $this->registerElement($group)
```
lippserd, line 101: This is just a personal thing, but I think the following looks better:

```
        $this
            ->registerElement($group)
            ->assembleGroup($group, $addElement ?? null, $removeElement ?? null)
            ->addHtml($group);

File: src/FormElement/Collection.php [hunk 9]
```diff
@@ -0,0 +1,143 @@
...
+
+        return $group;
+    }
+
+    /**
+     * @param string     $type
+     * @param string     $name
+     * @param array|null $attributes
+     *
+     * @return Collection
```
lippserd, line 113: Please use `@return $this` for fluent methods everywhere.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,143 @@
...
+     */
+    public function setAddElement(string $type, string $name, array $attributes = null): Collection
```
lippserd, line 115: Is `...): self` supported since PHP version 7.2? If so, I would use that.

lippserd (changes requested):

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+    protected $onAssembleGroup;
+
+    /** @var array */
+    protected $addElement;
+
+    /** @var array */
+    protected $removeElement;
+
+    /** @var string */
+    protected const FIELDSET_CSS_CLASS = 'form-element-collection';
```
lippserd, line 21: Please move this assignment to the beginning and rename it to `GROUP_CSS_CLASS`. If we ever replace the fieldset element, we are prepared.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+
+    /** @var string[] */
```
lippserd, line 23: You can remove the PHPDoc as the property is inherited.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+    protected function assembleGroup($group, $addElement, $removeElement): self
+    {
+        if (is_callable($this->onAssembleGroup)) {
+            call_user_func($this->onAssembleGroup, $group, $addElement, $removeElement);
+        }
+
+        return $this;
+    }
+
+    protected function assemble()
```
lippserd, line 42: Please move this method to the end of the protected methods block.

The general order should be public, protected, private, with consts and static first and overridden properties or methods after those the class defines.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * @param string     $type
+     * @param string     $name
+     * @param array|null $attributes
+     *
+     * @return $this
+     */
+    public function setAddElement(string $type, string $name, array $attributes = null): self
```
lippserd, line 105: Please move this method up. See above.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * @param string     $type
+     * @param string     $name
+     * @param array|null $attributes
+     *
+     * @return $this
+     */
+    public function setRemoveElement(string $type, string $name, array $attributes = null): self
```
lippserd, line 123: Please move this method up. See above.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+            $lastKey = $values ? key(array_slice($values, -1, 1, true)) + 1 : 0;
+            $this->addGroup($lastKey);
+        }
+    }
+
+    protected function addGroup($key): Fieldset
+    {
+        $group = new Fieldset(
+            $key,
+            Attributes::create(['class' => self::FIELDSET_CSS_CLASS])
```
lippserd, line 71: Use `static::`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+        );
+
+        if (isset($this->addTrigger['type']) && isset($this->addTrigger['name'])) {
```
lippserd, line 74: Use `if ($this->addElement !== null) {`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+            $addElement = $this->createElement(
+                $this->addTrigger['type'],
+                $this->addTrigger['name'],
+                $this->addTrigger['attributes']
+            );
+        }
+
+        if (isset($this->removeTrigger['type']) && isset($this->removeTrigger['name'])) {
```
lippserd, line 82: Use `if ($this->removeElement !== null) {`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+
+class Collection extends Fieldset
+{
+    /** @var callable */
+    protected $onAssembleGroup;
+
+    /** @var array */
+    protected $addElement;
```
lippserd, line 15: Make sure that you're using `addElement` everywhere. You also still use `addTrigger`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+
+    /** @var array */
+    protected $removeElement;
```
lippserd, line 18: Make sure that you're using `removeElement` everywhere. You also still use `removeTrigger`.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+
+        if (isset($this->removeTrigger['type']) && isset($this->removeTrigger['name'])) {
+            $removeElement = $this->createElement(
+                $this->removeTrigger['type'],
+                $this->removeTrigger['name'],
+                $this->removeTrigger['attributes']
+            );
+        }
+
+        $this
```
lippserd, line 90: Please use the following:

```
        $this
            ->assembleGroup($group, $addElement ?? null, $removeElement ?? null)
            ->addElement($group);
```

Else the `decorate()` call is missing.

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+
+    protected function assemble()
+    {
+        $this->addHtml(new HtmlElement('legend', null, new Text($this->getLabel())));
+
+        $values = $this->getPopulatedValues();
+
+        $valid = true;
+        foreach ($values as $key => $items) {
+            if ($this->removeElement !== null && isset($items[0][$this->removeElement['name']]) && count($values) > 1) {
```
lippserd, line 50: Do we really need the `count($values) > 1` protection? The element could be removed and an empty one would be added with the current code. Also how does `$items[0][$this->removeElement['name']]` work? Why the zero index?

src/FormElement/Collection.php
```diff
@@ -0,0 +1,133 @@
...
+            $group = $this->addGroup($key);
+
+            if (empty($group->getValue())) {
+                $valid = false;
+            }
+        }
+
+        if ($valid) {
+            $lastKey = $values ? key(array_slice($values, -1, 1, true)) + 1 : 0;
+            $this->addGroup($lastKey);
```
lippserd, line 63: Could this be just `$this->addGroup($key ? ($key + 1): 0);`?

TAINCER: The problem is getting the last key in the first place. Since this is not the foreach anymore. Functionality wise this should act like [array_key_last](https://www.php.net/manual/en/function.array-key-last.php), which is only available in 7.3 and up.

src/FormElement/Collection.php
TAINCER: With `addElememt()` the items added from assembleGroup are no longer added. I tried adding `->decorate($group)` before the `addHtml` call, and the output was the same in my test cases, also in the Unit Tests.

bobapple:
@cla-bot check

lippserd (changes requested):
Please rebase with master and add proper class and method PHPDoc. Also example usage for the element would be nice.

PR #73, opened 2022-08-29
SelectElementTest: test label change

Thomas-Gelf:


Thomas-Gelf:
Hint: please merge before #70, this should protect against a small issue that would otherwise break changing option labels

nilmerg:
Thanks.. but you could have also left a review there pointing this out and demanding this test. With the benefit of a learning effect.

PR #74, opened 2022-09-05
Replace SubFormElement with FieldsetElement

lippserd:


lippserd:
@TAINCER Please rewrite the tests to verify that the array notation works both in terms of rendering and value processing. Also make sure that it is still possible to get elements with their original names.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+    <option selected value="1">One</option>
+    <option value="2">Two</option>
+  </select>
+</fieldset>
+HTML;
+
+        $this->assertHtml($expected, $this->fieldset);
+    }
+
+    public function testTwoDimension(): void
```
lippserd, line 63: Please rename to `testTwoDimensions()`. And maybe you should name the fieldsets outer and inner, so that it's obvious in the expected HTML where they need to be placed.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+    private $fieldset2;
+
+    public function setup(): void
+    {
+        $this->name = 'test_fieldset';
+        $this->fieldset = new Fieldset($this->name);
+        $this->fieldset2 = new Fieldset($this->name . '2');
+    }
+
+    public function testConstruct(): void
```
lippserd, line 26: You are basically testing your test. This can be removed.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+    {
+        $this->assertInstanceOf(Fieldset::class, $this->fieldset);
+    }
+
+    public function testNameMethod(): void
```
lippserd, line 31: You are basically testing your test. This can be removed.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+    /** @var string */
+    private $name;
+
+    /** @var Fieldset */
+    private $fieldset;
+
+    /** @var Fieldset */
+    private $fieldset2;
+
+    public function setup(): void
```
lippserd, line 19: I would not prepare the objects in `setUp()`, but in each test case, so that it is immediately obvious how the test works and what it tests. I would use `setUp()` to prepare the environment needed for execution, such as setting up fixtures, but that is not necessary here. Also, not all objects created in `setUp()` are used in every test, so anyone looking at the tests will have to figure out what is used where.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+HTML;
+
+        $this->assertHtml($expected, $this->fieldset);
+    }
+
+    // Test with for loop
+    public function testSubFieldsetWithIterator(): void
+    {
+        $elementValues = [
+            ['timescale' => 10],
```
lippserd, line 101: Try to use the same options for all tests, i.e. `[null => 'Nothing', 1 => 'One', 2 => 'Two']`. As I write this, I think it makes sense to use `[null => 'Nothing', 1 => 'One', 'two' => 'Two']` (note the key `two` instead of `2`) for all tests. This way we always test different key types.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+    </select>
+  </fieldset>
+</fieldset>
+HTML;
+
+        $this->assertHtml($expected, $this->fieldset);
+    }
+
+    // Test with for loop
+    public function testSubFieldsetWithIterator(): void
```
lippserd, line 98: Please rename to `testMultipleDimensions()`. You only use the iterator to set up the elements, but the name of the test implies that there is a method in the element that does something with an iterator.

tests/FormElement/FieldsetTest.php
lippserd: I also think it makes sense to mix the types string and int for the field names here.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,173 @@
...
+            $elementValues
+        ]);
+
+        foreach ($elementValues as $key => $value) {
+            $fieldset = new Fieldset($key);
+
+            $fieldset->addElement('select', 'timescale', [
+                'value'   => $value['timescale'],
+                'options' => [
+                    [10 => 10, 20 => 20, 30 => 30]
```
lippserd, line 116: Please adjust the options here as well.

tests/FormElement/FieldsetTest.php
```diff
@@ -0,0 +1,146 @@
...
+namespace ipl\Tests\Html\FormElement;
+
+use ipl\Html\FormElement\Fieldset;
+use ipl\Tests\Html\TestCase;
+
+class FieldsetTest extends TestCase
+{
+    private const FIELDSET_NAME = "test_fieldset";
+
+    private const FIELDSET_OPTIONS = [
```
lippserd, line 12: The considerations from above apply here too, but since it defines select options (not fieldset options) that have to be tested, I'm ok with this constant if the name reflects what is expected, e.g. `SELECT_OPTIONS_TO_TEST`.

lippserd (commented):
I don't expect any more changes than the ones just requested, so after that please squash ll your commits belonging to `tests/FormElement/FieldsetTest.php` into one "Test Fieldset" commit.

src/FormElement/FieldsetElement.php
```diff
@@ -0,0 +1,115 @@
...
+            /**
+             * We don't change the {@see BaseFlement::$name} property of the element,
+             * otherwise methods like {@see FormElements::populate() and {@see FormElements::getElement()} would fail,
+             * but only change the name attribute to nest the names.
+             */
+            return sprintf(
+                '%s[%s]',
+                $this->getValueOfNameAttribute(),
+                $element->getName()
+            );
```
yhabteab, line 85: This won't work correctly for select elements with multiselect capability.

E.g:
```php
$fieldset->->addElement('select', 'days[]', [...]);
```

Results:
Bildschirmfoto 2022-10-24 um 12 39 02

lippserd: This is to be expected. We do not support array syntax specified by name. The file and select element must have the name specified without the brackets and the multiple attribute must be set to true.

yhabteab (dismissed):
Bildschirm­foto 2022-11-09 um 13 57 18

lippserd:
I think it's because cloning just doesn't work. But that's not something that this PR introduces.

lippserd:
> I think it's because cloning just doesn't work. But that's not something that this PR introduces.

Or did you spot something wrong here in the code @yhabteab?

yhabteab:
> Or did you spot something wrong here in the code @yhabteab?

I would say yes :) Instead of doing so, why couldn't the concrete form set the default decorator of the element?

https://github.com/Icinga/ipl-html/blob/91a5932104f2afbda7c896c01bfca157c7367f8c/src/FormElement/FieldsetElement.php#L60-L71

So, I've just removed the `prependWrapper()` and added this to the job config form and it works.
```php
$crontab>setDefaultElementDecorator(new IcingaFormDecorator());
```

lippserd:
> > Or did you spot something wrong here in the code @yhabteab?
> 
> I would say yes :) Instead of doing so, why couldn't the concrete form set the default decorator of the element?

I wouldn't expect a default decorator to have to be set multiple times.

Can you please try this [fix](https://github.com/Icinga/ipl-html/commit/cf543a393e5f3a68ec93030c3a076c9e1b1a9d57) please.

yhabteab:
> I wouldn't expect a default decorator to have to be set multiple times.

Its not about adding the decorator multiple times or not, but with my version we wouldn't have to clone things as we are instantiating our own decorator instance.

> Can you please try this [fix](https://github.com/Icinga/ipl-html/commit/cf543a393e5f3a68ec93030c3a076c9e1b1a9d57) please.

Bildschirm­foto 2022-11-10 um 10 08 40

lippserd:
> Its not about adding the decorator multiple times or not, but with my version we wouldn't have to clone things as we are instantiating our own decorator instance.

The whole point is that I don't want library users to have to set decorators for fieldsets if the form already provides a decorator.

Can you please try the latest commit with the cloning patch applied?

src/FormElement/FieldsetElement.php
```diff
@@ -0,0 +1,115 @@
...
+use ipl\Html\Contract\FormElement;
+use ipl\Html\Contract\FormElementDecorator;
+use ipl\Html\Contract\Wrappable;
+use LogicException;
+use ReflectionProperty;
+
+class FieldsetElement extends BaseFormElement
+{
+    use FormElements {
+        getValue as private getElementValue;
```
yhabteab, line 16:
```suggestion
        FormElements::getValue as private getElementValue;
```

lippserd: Why?

yhabteab: Bildschirm­foto 2022-11-10 um 12 17 16

lippserd: And what does PHPStorm say? It's not underlined for me and it's not a PSR-x violation if I'm not mistaken.

yhabteab: See here the last example: https://www.php-fig.org/psr/psr-12/#42-using-traits

lippserd: 🐘

tests/FormElement/FieldsetElementTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+    public function testOneDimension(): void
+    {
+        $fieldset = (new FieldsetElement('test_fieldset'))
+            ->addElement('select', 'test_select', ['options' => self::SELECT_OPTIONS_TO_TEST])
+            ->populate(['test_select' => 1]);
+
+        $expected = <<<'HTML'
+<fieldset name="test_fieldset">
+  <select name="test_fieldset[test_select]" value="1">
```
nilmerg, line 43: This will fail once #88 is merged. (`value` is not rendered anymore here)

Consider to use a different element. (A `select` isn't required here, is it?)

src/FormElement/FieldsetElement.php
```diff
@@ -0,0 +1,96 @@
...
+                $valid = false;
+            }
+        }
+
+        $this->valid = $valid;
+
+        return $this;
+    }
+
+    public function prependWrapper(Wrappable $wrapper)
```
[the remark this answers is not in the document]
lippserd: Nope, good catch.

src/FormElement/FieldsetElement.php
```diff
@@ -0,0 +1,96 @@
...
+        return $this;
+    }
+
+    public function getValueAttribute()
+    {
+        // Fieldsets do not have the value attribute.
+        return null;
+    }
+
+    public function validate()
```
nilmerg, line 46: Please rebase with master and remove this. It's not required anymore since #94

lippserd:
> This needs still a way to forward the validators messages to the base form, because this currently swallows the validators messages.

Good catch. Do you have an idea how to solve that?

yhabteab:
> > This needs still a way to forward the validators messages to the base form, because this currently swallows the validators messages.
> 
> Good catch. Do you have an idea how to solve that?

It's not this element's fault. This only happens with undecorated elements, i. e. if the element is properly added and decorated, the messages will be rendered as well.

nilmerg (approved):

nilmerg (approved):

lippserd:
@TAINCER Please rebase.

nilmerg (approved):

lippserd:
Wait before merging please. I think some commits can be squashed.

lippserd:
> Wait before merging please. I think some commits can be squashed.

Ready.

nilmerg (approved):

PR #76, opened 2022-09-21
FormElements: Don't return empty values when default value is provided

yhabteab:
When calling `getPopulatedValue()` with a default value, it sometimes returns `null/""` of the populated value and ignores the provided default value. This fix ensures that the default value is always returned if the populated value doesn't have a valid value.

### DIFF
--- a/src/FormElement/FormElements.php
+++ b/src/FormElement/FormElements.php
@@ -332,9 +332,14 @@ trait FormElements
      */
     public function getPopulatedValue($name, $default = null)
     {
-        return isset($this->populatedValues[$name])
-            ? $this->populatedValues[$name][count($this->populatedValues[$name]) - 1]
-            : $default;
+        if (isset($this->populatedValues[$name])) {
+            $value = $this->populatedValues[$name][count($this->populatedValues[$name]) - 1];
+            if (! empty($value)) {
+                return $value;
+            }
+        }
+
+        return $default;
     }
 
     /**


nilmerg:
I wouldn't call an empty value *invalid*. Why do you need this?

yhabteab:
> I wouldn't call an empty value _invalid_. Why do you need this?

When calling this method as follows:`$this->getPopulatedValue('every_months', 1)`, I would expect the method to return 1 if the populated value is empty/null. Having an input number and submitting the form without typing anything, the populated value is just `""` and I want to overwrite it with the default value.

nilmerg:
The thing is, this way you can't differentiate anymore between not sent / emptied. Though, I'm not yet sure this is important...

nilmerg:
So, after thinking about this again now I think this should be closed. Why?

The empty string is still passed on to the element. So a `$form->getValue('every_months', 1)` still returns it. As does the result of `$form->getValues()` include it. And this must not change and `$form->getPopulatedValue()` shouldn't behave differently.

As I hinted earlier, this is about not sent / emptied. It's a classic problem while working with forms. Most of the time there is no difference between not sent / emptied. For example in the case of configuration: Is the port set or not? Is a password set or should the connection be anonymous? The port may default to an official default as in the case of MySQL, the password however just should be not set and if one was set previously it should be cleared. Though, in both cases it doesn't matter if the browser transmitted a value for the element or if the user just didn't type in a value. The result is that in both cases, nothing gets stored.

Though, then there may be a select element with options like this:

```
[
    ''  => 'Please choose',
    'a' => 'Option A',
    'b' => 'Option B'
]
```

For a select there's always a value transmitted by the browser. There can't be nothing chosen. If the user leaves the select at `Please Choose` a call to `$form->getValues()` or `$form->getValue('select')` after the user submitted the form results in the empty string as expected.

By disabling the `Please Choose` option the user is obligated to choose something else. Though, if the element is not required, the browser just doesn't transmit the value anymore. A call to `$form->getValues()` doesn't include a value for the select and a call to `$form->getValue('select')` returns `null`.

I wouldn't expect `null` as result in both cases. No matter if calling `$form->getValue()` or `$form->getPopulatedValue()`.

--

So, I hope you see why not sent / emptied matters and that `$form->getValue()` and `$form->getPopulatedValue()` should return the same result.

Your example `$this->getPopulatedValue('every_months', 1)` should be `$this->getPopulatedValue('every_months') ?: 1` if you don't consider the empty string as valid value.

PR #80, opened 2022-10-05
CheckboxElement::setValue(): Use loose comparison if value is numeric

sukhwinder33445:
When checkedValue/uncheckedValue is a number, php casts it to string and strict check returns false


lippserd:
`getCheckedValue()` returns `string`. I don't expect anything else than string to work then. If you override those values, make sure those are strings.

sukhwinder33445:
> `getCheckedValue()` returns `string`. I don't expect anything else than string to work then. If you override those values, make sure those are strings.

Got it, thanks.

PR #86, opened 2022-10-31
Form: Only validate elements that have a value

### LINKED ISSUE #85: Form validation fails for non-required but empty elements when a validator is specified
The form currently [validates each element](https://github.com/Icinga/ipl-html/blob/949054cad5b3fa02ed2fd6dfb3091750fe8c6235/src/Form.php#L312) with given validators, which causes validation to fail due to empty elements that are not required.

## Expected Behavior
Non-required + empty elements do not need to be validated.

sukhwinder33445:
fixes #85

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -308,7 +308,10 @@ class Form extends BaseHtmlElement
                 $element->addMessage('This field is required');
                 $valid = false;
                 continue;
+            } elseif (! $element->hasValue()) {
+                continue;
             }
+
             if (! $element->isValid()) {
                 $valid = false;
             }
@@ -327,7 +330,9 @@ class Form extends BaseHtmlElement
     public function validatePartial()
     {
         foreach ($this->getElements() as $element) {
-            $element->validate();
+            if ($element->hasValue()) {
+                $element->validate();
+            }
         }
 
         return $this;


src/Form.php
```diff
@@ -308,7 +308,10 @@ public function validate()
                 $element->addMessage('This field is required');
                 $valid = false;
                 continue;
+            } elseif (! $element->hasValue()) {
```
lippserd, line 311: Instead of this `if ... elseif ...` construct with duplicate code (`$element->hasValue()` and `continue`) twice I suggest the following:

```
-            if ($element->isRequired() && ! $element->hasValue()) {
-                $element->addMessage('This field is required');
-                $valid = false;
+            if (! $element->hasValue()) {
+                if ($element->isRequired()) {
+                    $element->addMessage('This field is required');
+                    $valid = false;
+                }
+
                 continue;
             }
+
             if (! $element->isValid()) {
                 $valid = false;
             }
```

lippserd (changes requested):

nilmerg:
I'd like to refer here to Zend Form's `allowEmpty` flag. It's default `true` there and has the same effect as this change. Though, unlike here, it's adjustable. Set to `false`, empty values are still passed on to validators. We rely on this in Icinga DB Web for example. There the Redis configuration is only valid if both a TLS certificate and a key are defined. Both options can only be empty or set at the same time. This is validated with a callback validator and represents essentially a conditional `required` flag.

It may be possible to achieve the same by actually setting the `required` flag based on a condition. Though, by doing this it bypasses the validation mechanics and hence has some disadvantages. For one, you can't nicely provide a contextual error message. You also can't chain it with other validators. It's also chronologically wrong as it *validates* during assembly.

So, in my opinion this should also be adjustable.

lippserd:
`allowEmpty` sounds good to me but I would let it default to `false`.

nilmerg:
Could you please be more elaborate on why `false` as default?

lippserd (changes requested):
Please add a new property `validateEmpty` to the validators which is set to `false` by default (+ getter/setter). The `CallbackValidator` should have this property set to `true`. The validation chain must be adjusted to take the new property into account.

nilmerg:
The linked issue is now resolved by the change in ipl-validator.

PR #88, opened 2022-11-08
Enhance select element

### LINKED ISSUE #84: Disabled select element has null as value on submit
When a form is submitted having the selector element disabled, the value of the select element is null.

## Expected Behavior
The value should be the default value.

sukhwinder33445:
## Blocked by

- [x] https://github.com/Icinga/ipl-validator/pull/12
- [x] https://github.com/Icinga/ipl-html/pull/70

fixes #84

TODO:

- [x] Fix that test `testSelectingDisabledOptionIsNotPossible()` fails because of cached `isValid()`.

### DIFF
--- a/src/FormElement/SelectElement.php
+++ b/src/FormElement/SelectElement.php
@@ -29,33 +29,6 @@ class SelectElement extends BaseFormElement
     /** @var array|string */
     protected $value;
 
-    public function getValueAttribute()
-    {
-        // select elements don't have a value attribute
-        return null;
-    }
-
-    public function getNameAttribute()
-    {
-        $name = $this->getName();
-
-        return $this->isMultiple() ? ($name . '[]') : $name;
-    }
-
-    /**
-     * Get the value of the element
-     *
-     * Returns `array` when the attribute `multiple` is set to `true`, `string` otherwise
-     */
-    public function getValue()
-    {
-        if ($this->isMultiple()) {
-            return parent::getValue() ?? [];
-        }
-
-        return parent::getValue();
-    }
-
     /**
      * Get the option with specified value
      *
@@ -112,6 +85,33 @@ class SelectElement extends BaseFormElement
         }
 
         return $this;
+    }
+
+    /**
+     * Get the value of the element
+     *
+     * Returns `array` when the attribute `multiple` is set to `true`, `string` otherwise
+     */
+    public function getValue()
+    {
+        if ($this->isMultiple()) {
+            return parent::getValue() ?? [];
+        }
+
+        return parent::getValue();
+    }
+
+    public function getValueAttribute()
+    {
+        // select elements don't have a value attribute
+        return null;
+    }
+
+    public function getNameAttribute()
+    {
+        $name = $this->getName();
+
+        return $this->isMultiple() ? ($name . '[]') : $name;
     }
 
     /**
--- a/src/FormElement/SelectElement.php
+++ b/src/FormElement/SelectElement.php
@@ -2,12 +2,17 @@
 
 namespace ipl\Html\FormElement;
 
+use ipl\Html\Attributes;
+use ipl\Html\Common\MultipleAttribute;
 use ipl\Html\Html;
 use ipl\Validator\DeferredInArrayValidator;
 use ipl\Validator\ValidatorChain;
+use UnexpectedValueException;
 
 class SelectElement extends BaseFormElement
 {
+    use MultipleAttribute;
+
     protected $tag = 'select';
 
     /** @var SelectOption[] */
@@ -15,27 +20,29 @@ class SelectElement extends BaseFormElement
 
     protected $optionContent = [];
 
-    public function __construct($name, $attributes = null)
-    {
-        $this->getAttributes()->registerAttributeCallback(
-            'options',
-            null,
-            [$this, 'setOptions']
-        );
-        // ZF1 compatibility:
-        $this->getAttributes()->registerAttributeCallback(
-            'multiOptions',
-            null,
-            [$this, 'setOptions']
-        );
-
-        parent::__construct($name, $attributes);
-    }
+    /** @var array|string */
+    protected $value;
 
     public function getValueAttribute()
     {
         // select elements don't have a value attribute
         return null;
+    }
+
+    public function getNameAttribute()
+    {
+        $name = $this->getName();
+
+        return $this->isMultiple() ? ($name . '[]') : $name;
+    }
+
+    public function getValue()
+    {
+        if ($this->isMultiple()) {
+            return parent::getValue() ?? [];
+        }
+
+        return parent::getValue();
     }
 
     public function hasOption($value)
@@ -125,11 +132,29 @@ class SelectElement extends BaseFormElement
      */
     protected function isSelectedOption($optionValue)
     {
+        $value = $this->getValue();
+
+        if ($this->isMultiple()) {
+            if (! is_array($value)) {
+                throw new UnexpectedValueException(
+                    'Value must be an array when the `multiple` attribute is set to `true`'
+                );
+            }
+
+            return in_array($optionValue, $value, ! is_int($optionValue));
+        }
+
+        if (is_array($value)) {
+            throw new UnexpectedValueException(
+                'Value cannot be an array without setting the `multiple` attribute to `true`'
+            );
+        }
+
         return is_int($optionValue)
             // The loose comparison is required because PHP casts
             // numeric strings to integers if used as array keys
-            ? $this->getValue() == $optionValue
-            : $this->getValue() === $optionValue;
+            ? $value == $optionValue
+            : $value === $optionValue;
     }
 
     protected function addDefaultValidators(ValidatorChain $chain): void
@@ -154,5 +179,25 @@ class SelectElement extends BaseFormElement
     protected function assemble()
     {
         $this->addHtml(...array_values($this->optionContent));
+    }
+
+    protected function registerAttributeCallbacks(Attributes $attributes)
+    {
+        parent::registerAttributeCallbacks($attributes);
+
+        $attributes->registerAttributeCallback(
+            'options',
+            null,
+            [$this, 'setOptions']
+        );
+
+        // ZF1 compatibility:
+        $this->getAttributes()->registerAttributeCallback(
+            'multiOptions',
+            null,
+            [$this, 'setOptions']
+        );
+
+        $this->registerMultipleAttributeCallback($attributes);
     }
 }
--- a/src/FormElement/SelectElement.php
+++ b/src/FormElement/SelectElement.php
@@ -3,6 +3,8 @@
 namespace ipl\Html\FormElement;
 
 use ipl\Html\Html;
+use ipl\Validator\DeferredInArrayValidator;
+use ipl\Validator\ValidatorChain;
 
 class SelectElement extends BaseFormElement
 {
@@ -35,25 +37,6 @@ class SelectElement extends BaseFormElement
         return isset($this->options[$value]);
     }
 
-    public function validate()
-    {
-        $value = $this->getValue();
-        if (
-            $value !== null && (
-                ! ($option = $this->getOption($value))
-                || $option->getAttributes()->has('disabled')
-            )
-        ) {
-            $this->addMessage("'$value' is not allowed here");
-        } elseif ($this->isRequired() && $value === null) {
-            $this->addMessage('This field is required');
-        } else {
-            return parent::validate();
-        }
-
-        return false;
-    }
-
     public function deselect()
     {
         $this->setValue(null);
@@ -65,10 +48,6 @@ class SelectElement extends BaseFormElement
     {
         if ($option = $this->getOption($value)) {
             $option->getAttributes()->add('disabled', true);
-        }
-
-        if ($this->isSelectedOption($value)) {
-            $this->addMessage("'$value' is not allowed here");
         }
 
         return $this;
@@ -145,6 +124,25 @@ class SelectElement extends BaseFormElement
             // numeric strings to integers if used as array keys
             ? $this->getValue() == $optionValue
             : $this->getValue() === $optionValue;
+    }
+
+    protected function addDefaultValidators(ValidatorChain $chain): void
+    {
+        $chain->add(
+            new DeferredInArrayValidator(function (): array {
+                $possibleValues = [];
+
+                foreach ($this->options as $option) {
+                    if ($option->getAttributes()->get('disabled')->getValue()) {
+                        continue;
+                    }
+
+                    $possibleValues[] = $option->getValue();
+                }
+
+                return $possibleValues;
+            })
+        );
     }
 
     protected function assemble()
--- a/src/FormElement/SelectElement.php
+++ b/src/FormElement/SelectElement.php
@@ -26,13 +26,13 @@ class SelectElement extends BaseFormElement
     /** @var array Disabled select options */
     protected $disabledOptions = [];
 
-    /** @var array|string */
+    /** @var array|string|null */
     protected $value;
 
     /**
      * Get the option with specified value
      *
-     * @param string|int $value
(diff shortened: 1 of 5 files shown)


src/FormElement/SelectElement.php
```diff
@@ -30,29 +51,33 @@ public function __construct($name, $attributes = null)
...
-            parent::validate();
+        if ($this->isMultiSelect) {
+            return $this->getName() . '[]';
         }
 
-        return $this;
+        return $this->getName();
+    }
+
+    public function setValue($value)
```
lippserd, line 69: Why are you doing this?

It can be easily bypassed:

```php
(new SelectElement())->setValue('oops')->setAttribute('multiple', true);
```

The getter should return an array if `multiple` is `true`. Introducing dependencies in setters is not a good idea.

src/FormElement/SelectElement.php
```diff
@@ -124,20 +157,34 @@ protected function makeOption($value, $label)
...
+     * Whether the given option is a selected option
+     *
+     * @param $optionValue
+     *
+     * @return bool
+     */
+    protected function isSelectedOption($optionValue)
+    {
+        if ($this->isMultiSelect) {
+            return in_array($optionValue, (array)$this->getValue(), ! is_int($optionValue));
```
lippserd, line 178: Space missing after `(array)`.

tests/FormElement/SelectElementTest.php
```diff
@@ -209,6 +217,53 @@ public function testDefaultValueIsSelected()
...
+            'value'   => ['1', '5'],
+            'options' => [
+                null => 'Please choose',
+                '1'  => 'The one',
+                '4'  => 'Four',
+                '5'  => 'Hi five'
+            ]
+        ]);
+
+        $html = <<<HTML
```
lippserd, line 255: Please use nowdoc instead of heredoc everywhere.

src/FormElement/SelectElement.php
```diff
@@ -124,20 +157,34 @@ protected function makeOption($value, $label)
...
             $this->options[$value] = $option;
 
             return $this->options[$value];
         }
     }
 
+    /**
+     * Whether the given option is a selected option
+     *
+     * @param $optionValue
```
lippserd, line 171: Type missing.

lippserd (changes requested):
Since you are adjusting the entire class, please define the proper arguments and return types for all methods the class provides itself.

lippserd (changes requested):
1. Please document the trait.
2. Please test the trait.
3. Don't drop ZF1 compatibility [yet](https://github.com/search?q=org%3AIcinga+org%3Agipfl+user%3AThomas-Gelf+multioptions&type=Code). I haven't checked if any of this applies to IPL forms, but I wouldn't bother to check either. The separate PR was fine.
4. The whole logic for disabling options is questionable as it is not possible to control this via construct time attributes and option groups cannot be disabled or the entire element (which effectively means all options (and groups) are automatically disabled). And the funniest thing is, it's not used at all, see [here](https://github.com/search?q=org%3AIcinga+org%3Agipfl+user%3AThomas-Gelf+disableOption&type=Code), [here](https://github.com/search?q=org%3AIcinga+org%3Agipfl+user%3AThomas-Gelf+disableOptions&type=Code), [here](https://github.com/search?q=org%3AIcinga+org%3Agipfl+user%3AThomas-Gelf+makeOption&type=Code) and [here](https://github.com/search?q=org%3AIcinga+org%3Agipfl+user%3AThomas-Gelf+SelectOption&type=Code). So we have a semi-implemented feature that is not used at all and just presents complexity for further changes. I would remove the ability to disable options.
5. And while we're removing stuff anyway, please drop `deselect()`. Also [not used](https://github.com/search?l=PHP&q=org%3AIcinga+org%3Agipfl+user%3AThomas-Gelf+deselect&type=Code).

src/FormElement/SelectElement.php
```diff
@@ -121,25 +101,80 @@ protected function makeOption($value, $label)
...
+        });
+
+        $option->getAttributes()->registerAttributeCallback(
+            'disabled',
+            function () use ($option) {
+                return in_array($option->getValue(), $this->disabledOptions, true);
+            },
+            function () use ($option) {
+                $this->disabledOptions[] = $option->getValue();
+                $this->valid = false;
```
lippserd, line 118: Do not do that. For example, setting `required` after validation doesn't reset `valid` either. Overall I would say there are way too many places to watch out for. I would rather suggest not caching `valid` in general. We only call it once anyway. But that's not something this PR should address.

lippserd (changes requested):

lippserd (changes requested):

src/FormElement/SelectElement.php
```diff
@@ -2,144 +2,194 @@
...
-                    // numeric strings to integers if used as array keys
-                    ? $this->getValue() == $optionValue
-                    : $this->getValue() === $optionValue;
-            });
-            $this->options[$value] = $option;
-
-            return $this->options[$value];
-        }
+                foreach ($this->options as $option) {
+                    if ($option->getAttributes()->get('disabled')->getValue() === true) {
```
nilmerg, line 159: Please relax this. Someone might set `1` or `'disabled'` instead of `true`.

src/FormElement/SelectElement.php
```diff
@@ -2,144 +2,194 @@
...
+    public function getValue()
     {
-        $this->setValue(null);
+        $value = parent::getValue();
 
-        return $this;
+        return $this->isMultiple() ? (array) $value : $value;
     }
 
-    public function disableOption($value)
```
nilmerg, line 65: Please restore the possibility to disable options.

This test must pass:

```php
    /**
     * @depends testNullAndTheEmptyStringAreEquallyHandled
     */
    public function testDisablingOptionsIsWorking()
    {
        $form = new Form();
        $form->addElement('select', 'select', [
            'options' => ['' => 'Please choose', 'foo' => 'FOO', 'bar' => 'BAR'],
            'disableOptions' => [''],
            'required' => true,
            'value' => ''
        ]);

        $this->assertHtml(<<<'HTML'
<select name="select" required>
  <option value selected disabled>Please choose</option>
  <option value="foo">FOO</option>
  <option value="bar">BAR</option>
</select>
HTML
            ,
            $form->getElement('select')
        );
    }
```

src/FormElement/SelectElement.php
```diff
@@ -2,144 +2,194 @@
...
     {
-        if ($option = $this->getOption($value)) {
-            $option->getAttributes()->add('disabled', true);
-        }
-        if ($this->getValue() == $value) {
-            $this->valid = false;
-            $this->addMessage("'$value' is not allowed here");
-        }
+        $this->value = $value;
+        $this->valid = null;
```
nilmerg, line 79: This has been introduced based on my suggestion. It should ultimately let the test below pass. Though, for this the identification of selected options may also be adjusted alternatively.


```php

    public function testNullAndTheEmptyStringAreEquallyHandled()
    {
        $form = new Form();
        $form->addElement('select', 'select', [
            'options' => ['' => 'Please choose'],
            'value' => ''
        ]);
        $form->addElement('select', 'select2', [
            'options' => [null => 'Please choose'],
            'value' => null
        ]);

        $select = $form->getElement('select');
        $select2 = $form->getElement('select2');

        $this->assertInstanceOf(SelectOption::class, $select->getOption(''));
        $this->assertInstanceOf(SelectOption::class, $select2->getOption(null));
        $this->assertInstanceOf(SelectOption::class, $select->getOption(null));
        $this->assertInstanceOf(SelectOption::class, $select2->getOption(''));

        $this->assertTrue($select->isValid());
        $this->assertTrue($select2->isValid());

        $this->assertHtml(<<<'HTML'
<select name="select">
  <option value selected>Please choose</option>
</select>
HTML
            ,
            $select
        );
        $this->assertHtml(<<<'HTML'
<select name="select2">
  <option value selected>Please choose</option>
</select>
HTML
            ,
            $select2
        );
    }
```

nilmerg (changes requested):

src/FormElement/SelectElement.php
nilmerg: I've updated the test slightly. I prefer the alternative solution now (option identification) to make this non-distinction only an implementation detail.

```php
    public function testNullAndTheEmptyStringAreEquallyHandled()
    {
        $form = new Form();
        $form->addElement('select', 'select', [
            'options' => ['' => 'Please choose'],
            'value' => ''
        ]);
        $form->addElement('select', 'select2', [
            'options' => [null => 'Please choose'],
            'value' => null
        ]);

        /** @var SelectElement $select */
        $select = $form->getElement('select');
        /** @var SelectElement $select2 */
        $select2 = $form->getElement('select2');

        $this->assertNull($select->getValue());
        $this->assertNull($select2->getValue());

        $this->assertInstanceOf(SelectOption::class, $select->getOption(''));
        $this->assertInstanceOf(SelectOption::class, $select2->getOption(null));
        $this->assertInstanceOf(SelectOption::class, $select->getOption(null));
        $this->assertInstanceOf(SelectOption::class, $select2->getOption(''));

        $this->assertTrue($select->isValid());
        $this->assertTrue($select2->isValid());

        $select->setValue(null);
        $this->assertTrue($select->isValid());
        $select2->setValue('');
        $this->assertTrue($select2->isValid());

        $this->assertHtml(<<<'HTML'
<select name="select">
  <option value selected>Please choose</option>
</select>
HTML
            ,
            $select
        );
        $this->assertHtml(<<<'HTML'
<select name="select2">
  <option value selected>Please choose</option>
</select>
HTML
            ,
            $select2
        );
    }
```

src/FormElement/SelectElement.php
```diff
@@ -2,144 +2,219 @@
...
-        $this->optionContent = [];
-        foreach ($options as $value => $label) {
-            $this->optionContent[$value] = $this->makeOption($value, $label);
+        if ($optionValue === '') {
+            $optionValue = null;
         }
 
-        return $this;
+        if ($this->isMultiple()) {
+            return in_array($optionValue, $this->getValue(), ! is_int($optionValue));
```
nilmerg, line 155: `setValue([''])` isn't transformed to `setValue([null])`, so:

`in_array(null, [''], true) === false`

src/FormElement/SelectElement.php
```diff
@@ -2,144 +2,219 @@
...
+        $this->optionContent = [];
+        foreach ($options as $value => $label) {
+            $this->optionContent[$value] = $this->makeOption($value, $label);
+        }
+
+        return $this;
     }
 
-    public function validate()
+    public function setDisabledOptions(array $disabledOptions): self
```
nilmerg, line 66: Documentation please.

tests/FormElement/SelectElementTest.php
```diff
@@ -223,39 +236,117 @@ public function testSetValueSelectsAnOption()
...
+
+        $select->setValue(['1', 5]);
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Cannot set multiple values without setting the `multiple` attribute to `true`');
+
+        $select->render();
+    }
+
+    public function testSetArrayAsValueWithMultipleAttributeSetTheOptions()
```
nilmerg, line 297: Please add a case for `setValue([null])` and `setValue([''])`

nilmerg (changes requested):

src/FormElement/SelectElement.php
```diff
@@ -2,144 +2,227 @@
...
-        $this->optionContent = [];
-        foreach ($options as $value => $label) {
-            $this->optionContent[$value] = $this->makeOption($value, $label);
+        if ($optionValue === '') {
+            $optionValue = null;
         }
 
-        return $this;
+        if ($this->isMultiple()) {
+            return ($optionValue === null && array_filter($this->getValue()) === [])
```
nilmerg, line 162: Why `array_filter`?? This can only result in an empty array if `''` or `null` is *chosen*, exclusively. This check fails if another option is chosen.

nilmerg (changes requested):
Missing tests:

* `testSetOptionsResetsOptions()`
* `testOrderOfOptionsAndDisabledOptionsDoesNotMatter()`
* `testGetOptionReturnsPreviouslySetOption()`
* ```php
    public function testNullAndTheEmptyStringAreAlsoEquallyHandledWhileDisablingOptions()
    {
        $radio = new SelectElement('radio');
        $radio->setOptions([null => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions([null]);

        $this->assertTrue($radio->getOption(null)->getAttributes()->get('disabled')->getValue());

        $radio = new SelectElement('radio');
        $radio->setOptions(['' => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions(['']);

        $this->assertTrue($radio->getOption('')->getAttributes()->get('disabled')->getValue());

        $radio = new SelectElement('radio');
        $radio->setOptions([null => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions(['']);

        $this->assertTrue($radio->getOption(null)->getAttributes()->get('disabled')->getValue());
        $radio = new SelectElement('radio');
        $radio->setOptions(['' => 'Foo', 'bar' => 'Bar']);
        $radio->setDisabledOptions([null]);

        $this->assertTrue($radio->getOption('')->getAttributes()->get('disabled')->getValue());
    }
  ```
* ```php

    public function testGetOptionGetValueAndElementGetValueHandleNullAndTheEmptyStringEqually()
    {
        $radio = new SelectElement('radio');
        $radio->setOptions(['' => 'Foo']);
        $radio->setValue('');

        $this->assertNull($radio->getValue());
        $this->assertNull($radio->getOption('')->getValue());

        $radio = new SelectElement('radio');
        $radio->setOptions([null => 'Foo']);

        $this->assertNull($radio->getValue());
        $this->assertNull($radio->getOption(null)->getValue());
    }
  ```

src/FormElement/SelectElement.php
sukhwinder33445: Because it should not be possible to select the option `null` and some other options at the same time. $select->setValue(['', 5, 4]);` should mark only `5` and `4` as selected. 
I prefer to disable the '', null as default since they cannot be accepted as a valid value anyway.

nilmerg: Not possible? Only if disabled. If enabled, the option should of course be choosable. Either way, this decision is not this code's duty. This is the validator's duty. An text input also isn't empty if invalid after all.. (Btw. if disabled, the browser doesn't let the user choose the option anyway)

sukhwinder33445: Yes, got it. I have already added the changes after my comment.

src/FormElement/SelectElement.php
```diff
@@ -2,116 +2,129 @@
...
+        if (! empty($this->options)) {
+            foreach ($this->options as $option) {
+                $optionValue = $option->getValue();
 
-        return $this;
-    }
+                $option->setAttribute(
+                    'disabled',
+                    in_array($optionValue, $disabledOptions, ! is_int($optionValue))
+                    || ($optionValue === null && in_array('', $disabledOptions))
```
nilmerg, line 83: strict pls

nilmerg (approved):

src/FormElement/SelectElement.php
```diff
@@ -2,116 +2,134 @@
...
+     *
+     * @return SelectOption
+     *
+     * @throws InvalidArgumentException If no option with the specified value exists
+     */
+    public function getOption($value): SelectOption
     {
-        return isset($this->options[$value]);
+        if (! isset($this->options[$value])) {
+            throw new InvalidArgumentException(sprintf('There is no such option "%s"', $value));
```
nilmerg, line 44: This must not throw. It breaks compatibility way too hard and can't be worked around. Return null again if the option is unknown.

nilmerg (changes requested):

sukhwinder33445:
@cla-bot check

PR #89, opened 2022-11-08
SelectElement: Remove zf1 compatibility

sukhwinder33445:


sukhwinder33445:
Compatibility removed in https://github.com/Icinga/ipl-html/pull/88

PR #94, opened 2022-11-24
Move form validation utils to `FormElements`

yhabteab:


src/FormElement/FormElements.php
```diff
@@ -465,6 +468,66 @@ protected function decorate(FormElement $element)
...
+
+        return $this;
+    }
+
+    /**
+     * Validate all elements that have a value
+     *
+     * @return $this
+     */
+    public function validatePartial(): self
```
nilmerg, line 522: This must also call `ensureAssembled`

nilmerg: Also: remove the return type declaration. I don't want a breaking change just due to this.

nilmerg (changes requested):

src/FormElement/FormElements.php
```diff
@@ -465,6 +468,68 @@ protected function decorate(FormElement $element)
...
+     * @return $this
+     */
+    public function validate()
+    {
+        $this->ensureAssembled();
+
+        $valid = true;
+        foreach ($this->getElements() as $element) {
+            if ($element->isRequired() && ! $element->hasValue()) {
+                $element->addMessage('This field is required');
```
nilmerg, line 501: While you're at it, please translate this.

src/FormElement/FormElements.php
```diff
@@ -465,6 +468,68 @@ protected function decorate(FormElement $element)
         return $this;
     }
 
+    /**
+     * Get whether this form is valid
+     *
+     * {@see validate()} is called automatically if the form has not been validated before.
```
nilmerg, line 474: Use `self::validate()`, otherwise one might think this is a function.

src/FormElement/FormElements.php
```diff
@@ -465,6 +468,68 @@ protected function decorate(FormElement $element)
         return $this;
     }
 
+    /**
+     * Get whether this form is valid
```
nilmerg, line 472: Technically, this now returns whether all elements are valid or not.

nilmerg (changes requested):

src/FormElement/FormElements.php
```diff
@@ -19,6 +20,7 @@ trait FormElements
 {
     use Events;
     use Plugins;
+    use Translation;
```
nilmerg, line 23: Could you please check what happens if child classes also use this trait?

yhabteab: No errors, if that's what you're getting after.

yhabteab: Ah, wait! If you use both traits, i.e. `FormElements` and `Translation` in the same class, then there is a method collision, but if you use the translation trait in a child class of a class that uses `FormElements`, then it's fine.

nilmerg: > then there is a method collision

That's fine. Thanks.

nilmerg (approved):

PR #96, opened 2022-11-29
Cloning

lippserd:


src/Attributes.php
```diff
@@ -518,4 +511,78 @@ public function getIterator(): Traversable
...
         return new ArrayIterator($this->attributes);
     }
+
+    /**
+     * Rebind all {@see self::$callbacks} and {@see self::$setterCallbacks} that point to `$oldThis` to `$newThis`.
+     *
+     * @param object $oldThis
+     * @param object $newThis
+     *
+     * @return void
```
lippserd, line 521: You can omit `@return void` everywhere.

src/BaseHtmlElement.php
```diff
@@ -75,6 +76,9 @@ abstract class BaseHtmlElement extends HtmlDocument
     /** @var string Tag of element. Set this property in order to provide the element's tag when extending this class */
     protected $tag;
 
+    /** @var self */
```
lippserd, line 79: `@var $this`

src/Attributes.php
```diff
@@ -518,4 +510,60 @@ public function getIterator(): Traversable
     {
         return new ArrayIterator($this->attributes);
     }
+
+    /**
+     * Rebind all callbacks that point to `$oldThis` to `$newThis`.
```
lippserd, line 515: Please remove the full stop.

src/Attributes.php
```diff
@@ -518,4 +510,60 @@ public function getIterator(): Traversable
...
+    {
+        $this->rebindCallbacks($this->callbacks, $oldThis, $newThis);
+        $this->rebindCallbacks($this->setterCallbacks, $oldThis, $newThis);
+    }
+
+    /**
+     * Helper Method for {@see self::rebind()}
+     *
+     * Loops over all `$callbacks`, binds them to `$newThis` where `$oldThis` matches. The callbacks are
+     * modified directly on the `$callbacks` Reference.
```
lippserd, line 530: `reference` vs `Reference`

tests/CloneTest.php
```diff
@@ -0,0 +1,287 @@
...
+
+            if ($closureThis) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
```
lippserd, line 194: Remove the whole doc.

tests/CloneTest.php
```diff
@@ -0,0 +1,287 @@
...
+     * @param Attributes $attributes
+     * @param string     $name
+     *
+     * @return mixed
+     */
+    protected function getAttributeCallback(Attributes $attributes, string $name)
```
lippserd, line 200: Add `: callbable`.

tests/CloneTest.php
```diff
@@ -0,0 +1,287 @@
...
+    {
+        $callbacksProperty = new ReflectionProperty(get_class($attributes), 'callbacks');
+        $callbacksProperty->setAccessible(true);
+        $callbacks = $callbacksProperty->getValue($attributes);
+
+        if (isset($callbacks[$name])) {
+            return $callbacks[$name];
```
lippserd, line 207: Remove the `if` and the following `throw` so that the method implicitly throws an exception.

tests/CloneTest.php
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        throw new InvalidArgumentException();
+    }
+
+    /**
+     * @param callable $callback
+     *
+     * @return mixed|object|null
```
lippserd, line 216: Should only be `object|null` right?

tests/CloneTest.php
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    public function testCloningAttributes(): void
+    {
+        $original = Attributes::create([Attribute::create('class', 'class01')]);
+
+        $clone = clone $original;
+        $clone->setAttribute(Attribute::create('class', 'class02'));
+
+        $this->assertNotSame($original, $clone);
```
lippserd, line 168: Please test the attributes explicitly.

tests/CloneTest.php
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    public function testCloningSubmitElement(): void
+    {
+        $originalButton = new SubmitElement('exampleButton', ['class' => 'class01']);
+
+        $clonesButton = clone $originalButton;
+        $clonesButton->getAttributes()->setAttribute(Attribute::create('class', 'class02'));
+
+        $this->assertNotSame(
```
lippserd, line 100: Please test the attributes explicitly.

lippserd (commented):
Although this requires more effort, I would rather test the expected HTML instead of asserting that the output is not the same, as this could be caused by anything.

lippserd:
@TAINCER Please rebase.

lippserd:
@TAINCER @nilmberg has found that by keeping a reference to `$this` the object can only be destroyed at the end of the script. With a manual `unset()` call nothing happens. Since PHP 7.4, (WeakReference)[https://www.php.net/manual/en/class.weakreference.php] is available as a workaround. But since we are at PHP 7.2+, we have to solve this differently. Instead of keeping a reference to `$this`, we store the `spl_object_id()` and compare it later with the ID of the callback this.

tests/CloneTest.php
```diff
@@ -0,0 +1,260 @@
...
+            ]
+        ],
+        'SelectOption'  => 'test-value'
+    ];
+
+    public static function noopCallback()
+    {
+    }
+
+    public function testFormElementsCloning(): void
```
nilmerg, line 42: I don't quite understand the purpose of this test. The change proposed here fixes cloning of `Attributes` instances. These are part of `BaseHtmlElement`. But this test automatically tests any form element? Shouldn't it test any HTML element? Though, I think that just creating a html element with `Html::tag()` should be enough to test attribute cloning. There's no need for such a bloated test like this one here.

tests/CloneTest.php
```diff
@@ -0,0 +1,260 @@
...
+        ]);
+
+        $clone = clone $original;
+        $clone->addElement('submit', 'test_submit', [
+            'label' => 'Test Submit'
+        ]);
+        $clone->getAttributes()->setAttribute(Attribute::create('class', 'class02'));
+
+        $this->assertNotSame($original->render(), $clone->render());
+    }
```
nilmerg, line 155: Aren't these redundant? They check for nothing these form elements provide in particular. They're just re-iterations of `testFormElementsCloning` and `testCloningAttributes`.

tests/CloneTest.php
```diff
@@ -0,0 +1,260 @@
...
+use ipl\Html\BaseHtmlElement;
+use ipl\Html\Form;
+use ipl\Html\FormElement\InputElement;
+use ipl\Html\FormElement\SelectElement;
+use ipl\Html\FormElement\SubmitElement;
+use ReflectionException;
+use ReflectionFunction;
+use ReflectionProperty;
+
+class CloneTest extends TestCase
```
nilmerg, line 17: I'd now even go that far to propose making this test a full black box test. The use of reflection in here goes a bit too far for me. But I have no strong opinion here, I just think unit tests shouldn't be that complex. :smiley:

TAINCER: I tried to make the actual test cases as straightforward as possible in terms of logic, thus the Abstraction around Reflection, etc. Most fo the complex Logic is in the Helper methods, but I don't see how I would avoid this.

tests/CloneTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $attributes->registerAttributeCallback(
+            'test-closure-instance-element-get-property',
+            Closure::fromCallable([$this, 'getProperty'])
+        );
+    }
+}
+
+class CloneTest extends TestCase
+{
+    public function testElementCallbacksCloning(): void
```
nilmerg, line 51: This runs through because only the first parameter of `assertCallbacksFor` is actually in use.

tests/CloneTest.php
```diff
@@ -0,0 +1,168 @@
...
+    {
+        $element = new DummyElement();
+
+        $clone = clone $element;
+
+        $this->assertCallbacksFor($element, $this, null);
+        $this->assertCallbacksFor($clone, $this, $element);
+    }
+
+    public function testElementPropertyCloning()
```
nilmerg, line 61: This runs through because it's testing a PHP internal. @lippserd I was told you suggested this. I see **absolutely** no point in this test.

tests/CloneTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $element = new DummyElement();
+
+        $clone = clone $element;
+        $clone->customProperty = 'cloned_property';
+
+        $this->assertSame($element->getProperty(), 'original_property');
+        $this->assertSame($clone->getProperty(), 'cloned_property');
+    }
+
+    public function testCloningAttributes(): void
```
nilmerg, line 72: This runs through because the attribute object is overridden by using `setAttribute()`.

nilmerg (changes requested):

tests/CloneTest.php
nilmerg: I've pushed an update for `AttributesTest` with cases how I'd imagine them.

tests/CloneTest.php
lippserd: I did not suggest this because this PR, as discussed, is solely about attribute cloning.

tests/CloneTest.php
```diff
@@ -0,0 +1,168 @@
...
+
+use Closure;
+use ipl\Html\Attribute;
+use ipl\Html\Attributes;
+use ipl\Html\BaseHtmlElement;
+use ReflectionException;
+use ReflectionFunction;
+use ReflectionProperty;
+
+class DummyElement extends BaseHtmlElement
```
lippserd, line 13: Should be moved to `tests/lib`.

src/Attributes.php
```diff
@@ -365,29 +368,19 @@ public function setPrefix($prefix)
     /**
      * Register callback for an attribute
      *
-     * @param string   $name            Name of the attribute to register the callback for
-     * @param callable $callback        Callback to call when retrieving the attribute
-     * @param callable $setterCallback  Callback to call when setting the attribute
+     * @param string        $name           Name of the attribute to register the callback for
+     * @param callable|null $callback       Callback to call when retrieving the attribute
+     * @param callable|null $setterCallback Callback to call when setting the attribute
```
nilmerg, line 373: Please use `?type` instead of `type|null`. i.e. `?callable`.

src/Attributes.php
```diff
@@ -518,4 +511,66 @@ public function getIterator(): Traversable
...
+                    if (spl_object_id($callback[0]) === $oldThisId) {
+                        $newClosure = Closure::fromCallable($callback);
+                        $methodName = (new ReflectionFunction($newClosure))->name;
+
+                        $reflectedClosure = new ReflectionMethod($callback[0], $methodName);
+                        if ($reflectedClosure->isStatic()) {
+                            continue;
+                        }
+
+                        $callback = $newClosure->bindTo($newThis);
```
nilmerg, line 549:
```suggestion
                        $callback[0] = $newThis;
```

Or has this a disadvantage I'm not aware of?

tests/lib/CloningDummyElement.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace ipl\Tests\Html\lib;
```
nilmerg, line 3:
```suggestion
namespace ipl\Tests\Html\Lib;
```

tests/CloneTest.php
```diff
@@ -0,0 +1,157 @@
+<?php
+
+namespace ipl\Tests\Html;
+
+use Closure;
+use ipl\Html\Attribute;
+use ipl\Html\Attributes;
+use ipl\Tests\Html\lib\CloningDummyElement;
```
nilmerg, line 8:
```suggestion
use ipl\Tests\Html\Lib\CloningDummyElement;
```

tests/CloneTest.php
```diff
@@ -0,0 +1,157 @@
...
+            . '></p>';
+
+        $this->assertSame($originalHtml, $original->render());
+        $this->assertSame($firstCloneHtml, $firstClone->render());
+        $this->assertSame($secondCloneHtml, $secondClone->render());
+    }
+
+    public function testElementCallbacksCloning(): void
+    {
+        $element = new CloningDummyElement();
```
nilmerg, line 57:
```suggestion
        $element = new CloningDummyElement();
        $element->getAttributes();
```

Otherwise the test is useless. (Still runs through on master)

tests/CloneTest.php
```diff
@@ -0,0 +1,157 @@
...
+        $this->assertSame($original->get('class')->getValue(), 'class01');
+        $this->assertSame($clone->get('class')->getValue(), 'class02');
+    }
+
+    /**
+     * @param callable $callback
+     *
+     * @return object|null
+     * @throws ReflectionException
+     */
```
nilmerg, line 85: That docstring is still there. Why? Or, why do the other methods not also have one?

(Don't get me wrong: Please just remove this one)

nilmerg (changes requested):

src/Attributes.php
```diff
@@ -4,8 +4,11 @@
 
 use ArrayAccess;
 use ArrayIterator;
+use Closure;
 use InvalidArgumentException;
 use IteratorAggregate;
+use ReflectionFunction;
+use ReflectionMethod;
```
nilmerg, line 11: Unused now, remove it.

src/Attributes.php
```diff
@@ -365,29 +368,19 @@ public function setPrefix($prefix)
     /**
      * Register callback for an attribute
      *
-     * @param string   $name            Name of the attribute to register the callback for
-     * @param callable $callback        Callback to call when retrieving the attribute
-     * @param callable $setterCallback  Callback to call when setting the attribute
+     * @param string        $name           Name of the attribute to register the callback for
+     * @param ?callable     $callback       Callback to call when retrieving the attribute
+     * @param ?callable     $setterCallback Callback to call when setting the attribute
```
nilmerg, line 373: The whitespace changes are now not necessary anymore. Restore it to this:

```suggestion
     * @param string    $name           Name of the attribute to register the callback for
     * @param ?callable $callback       Callback to call when retrieving the attribute
     * @param ?callable $setterCallback Callback to call when setting the attribute
```

tests/CloneTest.php
```diff
@@ -0,0 +1,156 @@
+<?php
+
+namespace ipl\Tests\Html;
+
+use Closure;
+use ipl\Html\Attribute;
+use ipl\Html\Attributes;
+use ipl\Tests\Html\Lib\CloningDummyElement;
+use ReflectionException;
```
nilmerg, line 9: Unused now, remove it.

nilmerg (changes requested):

nilmerg (commented):
Woooh!

Now please squash everything into three commits:

1. 4f5106ed169b6159b236bfa288d827ce7654bdaa
2. Your CloneTest
3. Everything else

nilmerg (dismissed):
LGTM!

Now please push this on a different branch, rebased with master, and create a new PR. You're the author and @lippserd can't approve otherwise.

PR #97, opened 2022-11-30
Don't cache element validation state

yhabteab:
This change may break some div decorators and validate methods for the following modules as they use the `hasBeenValidated()`, `hasBeenValidateAndIsNotValid()` methods or have their own validation logic.

- https://github.com/Icinga/icingaweb2-module-reporting
   - `CompatDecorator` and `Flatpickr`
   - https://github.com/Icinga/icingaweb2-module-reporting/pull/168
- https://github.com/Icinga/icingaweb2-module-jira
   - `LegacyDecorator`
   - https://github.com/Icinga/icingaweb2-module-jira/pull/90
- https://github.com/gipfl/web
   - `MultiSelect` element
   - https://github.com/gipfl/web/pull/2

### DIFF
--- a/src/FormElement/BaseFormElement.php
+++ b/src/FormElement/BaseFormElement.php
@@ -30,9 +30,6 @@ abstract class BaseFormElement extends BaseHtmlElement implements FormElement, V
     /** @var bool Whether the element is required */
     protected $required = false;
 
-    /** @var null|bool Whether the element is valid; null if the element has not been validated yet, bool otherwise */
-    protected $valid;
-
     /** @var ValidatorChain Registered validators */
     protected $validators;
 
@@ -41,6 +38,13 @@ abstract class BaseFormElement extends BaseHtmlElement implements FormElement, V
 
     /** @var array Value candidates of the element */
     protected $valueCandidates = [];
+
+    /**
+     * @deprecated STOP: Do not use this property!!
+     *
+     * @var bool
+     */
+    protected $valid;
 
     /**
      * Create a new form element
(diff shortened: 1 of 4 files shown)


src/FormElement/BaseFormElement.php
```diff
@@ -262,16 +265,21 @@ public function onRegistered(Form $form)
...
-        $this->valid = $this->getValidators()->isValid($this->getValue());
+        $valid = $this->getValidators()->isValid($this->getValue());
         $this->addMessages($this->getValidators()->getMessages());
 
-        return $this;
+        return $valid;
     }
 
+    /**
+     * @inheritDoc
```
nilmerg, line 279: `@inheritDoc` is only needed in case you want the long description plus your new one. So here it is redundant. Please remove it.

src/FormElement/BaseFormElement.php
```diff
@@ -262,16 +265,21 @@ public function onRegistered(Form $form)
...
+     *
+     * @deprecated Use {@see self::hasMessages()} instead
```
nilmerg, line 281: hasMessages is no replacement for this. Just tell that it shouldn't be used at all.

nilmerg (changes requested):

nilmerg:
And please open PRs for reporting and jira regarding this change.

src/FormElement/BaseFormElement.php
```diff
@@ -153,23 +157,23 @@ public function setRequired($required = true)
...
     public function isValid()
     {
-        if ($this->valid === null) {
-            $this->validate();
+        if ($this->valid !== null) {
+            return $this->valid;
         }
 
-        return $this->valid;
+        return $this->validate();
```
nilmerg, line 164: validate may still return null if a child class overrides it with the old implementation in mind. Change it to this:

1. call validate, store result
2. return result if not null, this->valid otherwise

PR #104, opened 2023-01-10
Fix empty select value attribute

### LINKED ISSUE #103: `SelectElement` does not render empty `value` attribute
This causes the browser to send the label as value.

nilmerg:
fixes #103

### DIFF
--- a/src/FormElement/SelectOption.php
+++ b/src/FormElement/SelectOption.php
@@ -25,7 +25,7 @@ class SelectOption extends BaseHtmlElement
         $this->value = $value === '' ? null : $value;
         $this->label = $label;
 
-        $this->getAttributes()->registerAttributeCallback('value', [$this, 'getValue']);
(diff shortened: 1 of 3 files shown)


yhabteab (dismissed):
LFTM!

nilmerg:
I removed the return type hint from `getValueAttribute`. Please approve again, functionality didn't change.

yhabteab (approved):

lippserd:
> I removed the return type hint from `getValueAttribute`. Please approve again, functionality didn't change.

Why? How does it ever return `mixed`?

nilmerg:
Because https://github.com/Icinga/ipl-html/blob/master/src/FormElement/BaseFormElement.php#L323 and https://github.com/Icinga/ipl-html/blob/master/src/Attributes.php#L415-L422

lippserd:
But this method here does not return `mixed`.

PR #106, opened 2023-01-20
Attribute: Allow separator customization

nilmerg:
Some require a comma, e.g. the file input's accept attribute

### DIFF
--- a/src/Attribute.php
+++ b/src/Attribute.php
@@ -18,6 +18,9 @@ class Attribute
     /** @var string */
     protected $name;
 
+    /** @var string The separator used if value is an array */
+    protected $separator = ' ';
+
     /** @var string|array|bool|null */
     protected $value;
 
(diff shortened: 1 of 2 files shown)


lippserd:
Or `style`.

lippserd:
If we already know some specialties, do we want to support them right away?

nilmerg:
Hmm, the attribute doesn't know to which kind of element it belongs to. I don't want such an automatism just based on the name, you know :wink: 

The file input does such a thing then and I'm okay with it:

```
    public function __construct($name, $attributes = null)
    {
        $this->getAttributes()->get('accept')->setSeparator(', ');

        parent::__construct($name, $attributes);
    }
```

Could be improved, of course.

lippserd:
> Hmm, the attribute doesn't know to which kind of element it belongs to. I don't want such an automatism just based on the name, you know 😉

That is true.

lippserd (commented):
Shouldn't there also be `getSeparator()`?

nilmerg:
If you want it. I figured it will be not used and it's basically only an internal thing and of no use externally anyway.

lippserd:
> If you want it. I figured it will be not used and it's basically only an internal thing and of no use externally anyway.

I would add it. You never know if it is necessary to check the separator before adjusting it, or do other stuff based on the separator.

lippserd (approved):

PR #110, opened 2023-02-15
Clone attributes

lippserd:
Right now we have an attribute cloning use case where we [duplicate the primary submit button](https://github.com/Icinga/ipl-web/pull/104) and add it as the first element to ensure that it is used for implicit form submission. The submit button to be cloned can have attributes and callbacks that must also be present on the clone, such as `formaction`, `formmethod` and `formnovalidate`, so simply not using the original attributes is not an option.

Until [proper cloning](https://github.com/Icinga/ipl-html/pull/96) is implemented, it is imperative to re-register attribute callbacks for attributes where a callback exists and the clone provides a changed value.

### DIFF
--- a/src/Attributes.php
+++ b/src/Attributes.php
@@ -518,4 +518,11 @@ class Attributes implements ArrayAccess, IteratorAggregate
     {
         return new ArrayIterator($this->attributes);
     }
+
+    public function __clone()
+    {
+        foreach ($this->attributes as &$attribute) {
+            $attribute = clone $attribute;
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


lippserd:
See Icinga/ipl-web#104 for the duplicate submit button PR and #96 for proper cloning.

nilmerg (changes requested):
Please integrate the tests from #96 here.

lippserd:
> Please integrate the tests from #96 here.

Added my own test. The PR referenced contains two tests that basically do the same thing, but only change one attribute. Here we also test adding and removing attributes and cloning already cloned attributes.

nilmerg (approved):

PR #116, opened 2023-03-09
SubmitButtonElement: Fix that it is always pressed

### LINKED ISSUE #115: Multiple `SubmitButtonElement's` in one form do not work properly
## To Reproduce

1. Create a form with 2 or more `SubmitButtonElement's`.
2. Submit the form by clicking the 2nd or last submit button.
3. `$form->getPressedSubmitElement()` always returns the first submit button, even if it is not pressed.

nilmerg:
fixes #115

### DIFF
--- a/src/FormElement/SubmitButtonElement.php
+++ b/src/FormElement/SubmitButtonElement.php
@@ -2,13 +2,39 @@
 
 namespace ipl\Html\FormElement;
 
+use ipl\Html\Attributes;
 use ipl\Html\Contract\FormSubmitElement;
 
 class SubmitButtonElement extends ButtonElement implements FormSubmitElement
 {
     protected $defaultAttributes = ['type' => 'submit'];
 
-    protected $value = 'y';
+    /** @var string The value that's transmitted once the button is pressed */
+    protected $submitValue = 'y';
+
+    /**
+     * Get the value to transmit once the button is pressed
+     *
+     * @return string
+     */
+    public function getSubmitValue(): string
+    {
+        return $this->submitValue;
+    }
+
+    /**
+     * Set the value to transmit once the button is pressed
+     *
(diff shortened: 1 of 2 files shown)


lippserd:
Can this also be based on the label, as with `SubmitElement`? I don't like that the implementations differ so much. Also, if I understand it correctly, you would have to set different submit values when using multiple `SubmitButtonElement`s.

nilmerg:
> Can this also be based on the label, as with SubmitElement? I don't like that the implementations differ so much.

If we do this, there needs to be an option like Zend's `content`. Because the one advantage of `<button>` is that it differentiates between *value* and *label*.

> Also, if I understand it correctly, you would have to set different submit values when using multiple SubmitButtonElements.

Or a different name for each.

nilmerg:
## label as value

```php
$form->addElement('submitButton', 'foo', [
    'label' => 'Foo' // Will also be the submitted value
]);
$form->addElement('submitButton', 'bar', [
    'label' => 'bar',
    'content' => new Icon('bar') // Required to still support non-string-labels
]);

/* sometime later */
$form->getPressedSubmitElement()->getValue() === 'bar';
/* or */
$form->getElement('bar')->hasBeenPressed();
```

## submitValue as value

```php
$form->addElement('submitButton', 'abc', [
    'label' => 'ABC'
]);
$form->addElement('submitButton', 'cba', [
    'label' => new Icon('cba')
]);

/* sometime later */
$form->getPressedSubmitElement()->getName() === 'cba';
/* or */
$form->getElement('cba')->hasBeenPressed();
```

--

I personally like the *submitValue as value* approach better, because since I have to provide unique name anyway, I also check on the name and I am able to easily provide a non-string-label.

lippserd (approved):

PR #118, opened 2023-03-16
Form: Don't validate empty elements during partial validation

nilmerg:


lippserd (dismissed):

yhabteab (commented):
Such a use case doesn't work anymore. The Form validation fails but the messages are never rendered.

```php
$this->addElement('textarea', 'foo', [
    'validators'  => [
        new CallbackValidator(function ($value): bool {
            $this->getElement('foo')->addMessage($this->translate('Validation failed'));

            return false;
        })
    ]
]);
```

nilmerg:
Well, that's incorrect after all. The validator returns the validation result, so it should provide messages. A callback validator also passes itself as second parameter to the callback for exact this case.

yhabteab (commented):
I think we have to also add the same check [here](https://github.com/Icinga/ipl-html/pull/118/files#diff-00d692b66d8b2accfb0aca61fbdf883a9f7d9c7838b240b187aab8c5d266bfa5L46).

lippserd:
> I think we have to also add the same check [here](https://github.com/Icinga/ipl-html/pull/118/files#diff-00d692b66d8b2accfb0aca61fbdf883a9f7d9c7838b240b187aab8c5d266bfa5L46).

Difficult. We don't know whether it's partial validation. But yeah, it's inconsistent.

yhabteab:
However, the `gipfl` form elements do this as well. These could become broken with the new changes in iml-html.

nilmerg:
That's not a validator. And it isn't affected by the recent changes, as they reverted the behavior regarding messages, state and decorators back to what it was in the last release. If this is broken now, it was broken beforehand already, If I'm not mistaken.

lippserd:
> However, the `gipfl` form elements do this as well. These could become broken with the new changes in iml-html.

Why? That's `onSuccess()` where  full validation should've run beforehand.

lippserd:
> > Do you have a real usecase?
> 
> Yes! [Here](https://github.com/Icinga/ipl-web/blob/master/src/FormElement/ScheduleElement.php#L476-L480), though I will change it.

How is this broken?

yhabteab:
In case of an invalide cron expression the form validation will fail but you will never see the message.

lippserd (approved):
We had a brief discussion about the issues Yonas mentioned, and all but partial fieldset validation are related to the streamline validation PR, which is due to improper usage. A fix for partial fieldset validation will follow.

PR #119, opened 2023-03-17
SubmitButtonElement: Set the submit value by the `value` attribute

nilmerg:
Some forms still use `value` to define the submit value. It shouldn't make a difference to use it as submit value internally now, as the end result in HTML is the same.


lippserd (dismissed):
PR title and description should be adjusted though.

lippserd:
> PR title and description should be adjusted though.

And also commit title and description 😆

nilmerg:
Why? There's still a submit value that's not *the* value.

lippserd:
> 

Easy fix ;).

lippserd (approved):

PR #121, opened 2023-03-20
BaseFormElement: Only run validators if there's a value

nilmerg:
The counterpart for https://github.com/Icinga/ipl-validator/pull/29

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -47,6 +47,18 @@ class Form extends BaseHtmlElement
     protected $tag = 'form';
 
     /**
+     * Get whether the given value is empty
+     *
+     * @param mixed $value
+     *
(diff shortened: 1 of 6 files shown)


lippserd (dismissed):

yhabteab (dismissed):

nilmerg:
I've changed `FieldsetElement::hasValue()` so that it doesn't rely on what `getValue()` returns. This also fixes that nested fieldsets can't be empty once they have elements, as verified by the new test `testNestedFieldsetsAlsoOnlyHaveAValueIfOneOfTheirElementsHaveOne`.

yhabteab (approved):

lippserd (approved):

PR #123, opened 2023-06-19
Github Actions: Add PhpStan

sukhwinder33445:


src/HtmlDocument.php
```diff
@@ -328,8 +326,6 @@ public function remove(ValidHtml $html)
         $this->content = array_values($this->content);
 
         $this->reIndexContent();
-
-        return $this;
```
nilmerg, line 332: why is this removed and not added in the overrides?

src/FormDecorator/DdDtDecorator.php
```diff
@@ -53,7 +53,7 @@ public function getAttributes()
         if ($this->wrappedElement->hasBeenValidated() && ! $this->wrappedElement->isValid()) {
             $classes = $attributes->get('class');
             if (
-                empty($classes)
+                $classes->isEmpty()
```
nilmerg, line 56: There should rather `getValue` being called above. Otherwise `$classes` can never be a string or an array.

src/Attribute.php
```diff
@@ -325,4 +325,9 @@ public function renderValue()
     {
         return static::escapeValue($this->value, $this->separator);
     }
+
+    public function __toString()
```
nilmerg, line 329: Why is this needed? i.e. Where are attributes casted to string?

phpstan.neon
```diff
@@ -0,0 +1,44 @@
...
+        - '#Call to an undefined method ipl\\Html\\Contract\\FormElement\:\:ensureAssembled\(\)#' # DivDecorator -> Line 90
+
+
+        # methods with wrong phpDoc
+        - '#. should return .* but returns .#'
+        - '#. expects .* mixed given#'
+        - '#Property ipl\\Html\\HtmlDocument\:\:.#' #
+        - '#Property ipl\\Html\\Table\:\:.#' #
+        - '#Call to an undefined method ipl\\Html\\Contract\\Wrappable\:\:.#' # call to HtmlDocument methods on wrappable
+        - '#Parameter \#2 \$value of static method .#' # Attributes -> Line 43
```
nilmerg, line 44: Are all of these unfixable or did you just ignore them due to time constraints? If it's the latter, this should be in the baseline configuration.

nilmerg (changes requested):

nilmerg:
and please rebase

phpstan.neon
yhabteab: done

src/FormElement/FileElement.php
```diff
@@ -403,7 +403,10 @@ public function renderUnwrapped()
                 ])))->addHtml(new HtmlElement(
                     'span',
                     null,
-                    new Icon('remove'),
+                    Html::tag(
```
nilmerg, line 406: Mhm, we should use `HtmlElement` here as well. Introducing and using `Html` here just for this icon feels wrong.

nilmerg (changes requested):

src/FormElement/FileElement.php
nilmerg, line 1: Please amend Ravi's commit

nilmerg (approved):

PR #134, opened 2024-04-03
Make `FormElements` and `FormElement` generic

nilmerg:


src/Contract/FormElement.php
```diff
@@ -66,21 +68,24 @@ public function getName();
     /**
      * Get whether the element has a value
      *
+     * @phpstan-assert-if-true !null $this->getValue()
+     * @phpstan-assert-if-true non-empty-string $this->getValue()
+     * @phpstan-assert TValue $this->getValue()
```
nilmerg, line 73: still need to test this

src/FormElement/NumberElement.php
```diff
@@ -2,6 +2,7 @@
 
 namespace ipl\Html\FormElement;
 
+/** @extends InputElement<int> */
```
nilmerg, line 5: not quite correct, is it?

nilmerg:
Not merging. Doesn't work as I imagined. Possibly not the best candidate for generics either.

PR #141, opened 2025-02-14
Form: Treat `only whitespace` value as empty

sukhwinder33445:
fixes https://github.com/Icinga/icinga-notifications-web/issues/208

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -55,7 +55,7 @@ class Form extends BaseHtmlElement
      */
     public static function isEmptyValue($value): bool
     {
-        return $value === null || $value === '' || $value === [];
+        return $value === null || $value === [] || (is_string($value) && trim($value) === '');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Form.php
```diff
@@ -55,7 +55,7 @@ class Form extends BaseHtmlElement
      */
     public static function isEmptyValue($value): bool
     {
-        return $value === null || $value === '' || $value === [];
+        return $value === null || $value === [] || (is_string($value) && trim($value) === '');
```
ncosta-ic, line 58: This only catches common whitespaces like space, tabulator, line-breaks, ...

If you want to catch all whitespaces (including the unicode ones, you'd want this[^1] regex):
```php
(is_string($value) && preg_replace('/^[\pZ\pC]+|[\pZ\pC]+$/u', '', $value) === '');
```

[^1]: See https://stackoverflow.com/questions/4166896/trim-unicode-whitespace-in-php/4167053#4167053

sukhwinder33445: Can you please explain why this is necessary and what the use cases might look like?
Can you show some example strings that could lead to a problem?

This is a kind of special handling that in my opinion should not be added to form, and can be added as a validator to elements that explicitly require it.

ncosta-ic: Well, one could technically still create submissions with "empty looking" values.
But you're right. This case should better be catched with a validator

ncosta-ic (approved):
Looks good to me

nilmerg (approved):

PR #144, opened 2025-03-24
Check if actual checkbox value equals checkedValue

### LINKED ISSUE #143: Required checkboxes are not validated correctly on the server side
By default, in html you can't submit unchecked checkboxes if they are set to 'required'. In our forms, if you remove the 'required' tag in the browser's dev tools, you can submit a checkbox even if it's not checked. I think this shouldn't be possible to prevent submitting a form where the user's agreement is mandatory (terms and conditions for example).

jrauh01:
Override method `hasValue()` from `BaseFormElement` in `CheckboxElement`.

Fix #143

### DIFF
--- a/tests/FormElement/CheckboxElementTest.php
+++ b/tests/FormElement/CheckboxElementTest.php
@@ -3,9 +3,16 @@
 namespace ipl\Tests\Html;
 
 use ipl\Html\FormElement\CheckboxElement;
+use ipl\I18n\NoopTranslator;
+use ipl\I18n\StaticTranslator;
 
 class CheckboxElementTest extends TestCase
 {
+    protected function setUp(): void
+    {
+        StaticTranslator::$instance = new NoopTranslator();
+    }
+
     public function testRendersCheckValuesAsValueAttribute()
     {
         $checkbox = new CheckboxElement('test');
@@ -105,5 +112,14 @@ class CheckboxElementTest extends TestCase
             . '<input checked="checked" type="checkbox" name="test" value="checked">',
             $checkbox
         );
+    }
+
+    public function testCheckboxNotValidIfRequiredAndUnchecked()
+    {
+        $checkbox = new CheckboxElement('test');
+        $checkbox->setRequired();
+        $checkbox->setChecked(false);
+
+        $this->assertFalse($checkbox->isValid());
     }
 }
--- a/src/FormElement/CheckboxElement.php
+++ b/src/FormElement/CheckboxElement.php
@@ -121,4 +121,13 @@ class CheckboxElement extends InputElement
 
         return (new HiddenElement($this->getValueOfNameAttribute(), ['value' => $this->getUncheckedValue()])) . $html;
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


nilmerg (changes requested):
One thing to note is that an unchecked checkbox is now not validated anymore. But a quick search didn't yield any usage of an validator for checkboxes…

But, there's a test for this element, please add one covering this change.

nilmerg (approved):

tests/FormElement/CheckboxElementTest.php
```diff
@@ -106,4 +106,14 @@ public function testSetCheckValues()
...
         );
     }
+
+    public function testCheckboxNotValidIfRequiredAndUnchecked()
+    {
+        $checkbox = new CheckboxElement('test');
+        $checkbox->setRequired();
+        $checkbox->setChecked(false);
+
+        $this->assertSame('n', $checkbox->getValue());
```
nilmerg, line 116: Ah, didn't see the test results, but my first impression was anyway: This adds nothing to the case here, it's only about validity after all. So just remove this I'd say.

nilmerg (changes requested):

tests/FormElement/CheckboxElementTest.php
jrauh01: I think `$checkbox->setRequired()` has to be in the test, because only if the checkbox is 'required' the validation should fail.

jrauh01:
@nilmerg I think the test should now test the case correctly.

tests/FormElement/CheckboxElementTest.php
```diff
@@ -106,4 +106,13 @@ public function testSetCheckValues()
...
         );
     }
+
+    public function testCheckboxNotValidIfRequiredAndUnchecked()
+    {
+        $checkbox = new CheckboxElement('test');
+        $checkbox->setRequired();
+        $checkbox->setValue($checkbox->getUncheckedValue());
+
+        $this->assertFalse($checkbox->hasValue());
```
nilmerg, line 116: It's correct that you patched `hasValue()`, but the original problem was that the checkbox was valid, hence this should be verified. (It's also what the test name states)

Setting the value directly is uncommon for checkboxes. The native way is to *check* them, or not. In this case, it is necessary to ensure it is not checked. Remember, unit tests should preferably be black box tests.

```suggestion
        $checkbox = new CheckboxElement('test');
        $checkbox->setRequired();
        $checkbox->setChecked(false);

        $this->assertFalse($checkbox->isValid());
```

nilmerg (changes requested):

tests/FormElement/CheckboxElementTest.php
jrauh01: Yes, I agree with you. I had that before, but the test failed because of the `translate()` call on `null`.

$checkbox->isValid()
    $this->validate()
        $this->setMessages([$this->translate()])

nilmerg (approved):

nilmerg (approved):

PR #146, opened 2025-05-21
New decorator implementation

sukhwinder33445:
Note for myself:
- ~~Have a look at local branch `form-decoration`.~~
- Branch `enhance-decorator` introduces form decoration.

requires https://github.com/Icinga/ipl-html/pull/152

### DIFF
--- a/tests/FormDecorator/DecorationResultsTest.php
+++ b/tests/FormDecorator/DecorationResultsTest.php
@@ -3,35 +3,40 @@
 namespace ipl\Tests\Html\FormDecorator;
 
 use ipl\Html\FormDecorator\DecorationResults;
+use ipl\Html\FormDecorator\Transformation;
 use ipl\Html\Html;
-use ipl\Html\HtmlString;
 use ipl\Tests\Html\TestCase;
 
 class DecorationResultsTest extends TestCase
 {
+    protected DecorationResults $results;
+
+    public function setUp(): void
+    {
+        $this->results = new DecorationResults();
+    }
+
     public function testEmptyDecorationResultsRenderEmptyString(): void
     {
-        $this->assertSame('', (new DecorationResults())->render());
+        $this->assertSame('', $this->results->render());
     }
 
     public function testMethodSkipDecorators(): void
     {
-        $results = new DecorationResults();
-        $this->assertCount(0, $results->getSkipDecorators());
-        $this->assertSame([], $results->getSkipDecorators());
-
-        $results->skipDecorators('Decorator1');
-        $this->assertCount(1, $results->getSkipDecorators());
-        $this->assertSame(['Decorator1'], $results->getSkipDecorators());
-
-        $results->skipDecorators('Decorator2', 'Decorator3');
-        $this->assertCount(3, $results->getSkipDecorators());
-        $this->assertSame(['Decorator1', 'Decorator2', 'Decorator3'], $results->getSkipDecorators());
+        $this->assertSame([], $this->results->getSkipDecorators());
+        $this->assertSame(
+            ['Decorator1'],
+            $this->results->skipDecorators('Decorator1')->getSkipDecorators()
+        );
+        $this->assertSame(
+            ['Decorator1', 'Decorator2', 'Decorator3'],
+            $this->results->skipDecorators('Decorator2', 'Decorator3')->getSkipDecorators()
+        );
     }
 
     public function testMethodAppend(): void
     {
-        $results = (new DecorationResults())
+        $this->results
             ->append(Html::tag('div', ['class' => 'first'], 'First'))
             ->append(Html::tag('div', ['class' => 'second'], 'Second'))
             ->append(Html::tag('div', ['class' => 'third'], 'Third'));
@@ -42,12 +47,12 @@ class DecorationResultsTest extends TestCase
 <div class="third">Third</div>
 HTML;
 
-        $this->assertHtml($html, $results);
+        $this->assertHtml($html, $this->results);
     }
 
     public function testMethodPrepend(): void
     {
-        $results = (new DecorationResults())
+        $this->results
             ->prepend(Html::tag('div', ['class' => 'third'], 'Third'))
             ->prepend(Html::tag('div', ['class' => 'second'], 'Second'))
             ->prepend(Html::tag('div', ['class' => 'first'], 'First'));
@@ -58,12 +63,12 @@ HTML;
 <div class="third">Third</div>
 HTML;
 
-        $this->assertHtml($html, $results);
+        $this->assertHtml($html, $this->results);
     }
 
     public function testMethodWrap(): void
     {
-        $results = (new DecorationResults())
+        $this->results
             ->wrap(Html::tag('div', ['class' => 'wrapper-1']))
             ->wrap(Html::tag('div', ['class' => 'wrapper-2']))
             ->wrap(Html::tag('div', ['class' => 'wrapper-3']));
@@ -76,12 +81,12 @@ HTML;
 </div>
 HTML;
 
-        $this->assertHtml($html, $results);
+        $this->assertHtml($html, $this->results);
     }
 
     public function testAppendAfterWrap(): void
     {
-        $results = (new DecorationResults())
+        $this->results
             ->append(Html::tag('div', ['class' => 'first'], 'First'))
             ->wrap(Html::tag('div', ['class' => 'wrapper']))
             ->append(Html::tag('div', ['class' => 'second'], 'Second'));
@@ -93,11 +98,11 @@ HTML;
 <div class="second">Second</div>
 HTML;
 
-        $this->assertHtml($html, $results);
+        $this->assertHtml($html, $this->results);
     }
     public function testPrependAfterWrap(): void
     {
-        $results = (new DecorationResults())
+        $this->results
             ->append(Html::tag('div', ['class' => 'first'], 'First'))
             ->wrap(Html::tag('div', ['class' => 'wrapper']))
             ->prepend(Html::tag('div', ['class' => 'second'], 'Second'));
@@ -108,12 +113,12 @@ HTML;
   <div class="first">First</div>
 </div>
 HTML;
-        $this->assertHtml($html, $results);
+        $this->assertHtml($html, $this->results);
     }
 
     public function testMixed(): void
     {
-        $results = (new DecorationResults())
+        $this->results
             ->append(Html::tag('tag1'))
             ->wrap(Html::tag('tag2'))
             ->prepend(Html::tag('tag3'))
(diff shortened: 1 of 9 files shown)


src/FormDecorator/DecoratorChain.php
```diff
@@ -0,0 +1,229 @@
...
+            throw new InvalidArgumentException(sprintf(
+                "Can't load decorator '%s'. decorator unknown",
+                $name
+            ));
+        }
+
+        if (empty($options)) {
+            $decorator = new $class();
+        } else {
+            $decorator = new $class($options);
```
nilmerg, line 166: 1. The `Decorator` interface does not require a constructor argument.
2. `$options` are only handled by classes extending `BaseDecorator`.
3. However, the `DecoratorChain` is satisfied with classes implementing `Decorator`.
4. Repeat (goto 1.)

So why don't we do this instead:

Require implementors of `Decorator` to provide `Attributes`. (i.e. `getAttributes(): Attributes`) Use them here and pass `$options` to `$attributes->add()`.

src/FormDecorator/BaseDecorator.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Extend this class in order to provide concrete element decorator.
+ * When extending this class you can provide default css class with {@link static::CSS_CLASS}.
+ *
+ * Override {@link static::registerAttributeCallbacks()} method in order to register attribute callbacks.
+ */
+abstract class BaseDecorator implements Decorator
+{
+    /** @var string CSS class for decorator element */
+    protected const CSS_CLASS = 'form-element';
```
nilmerg, line 19: That is specific to the `HtmlTag` decorator, isn't it? Or rather, specific to the decorator usage of `HtmlTag`, that's a `div`, wrapping an entire decorated form element, that's not a fieldset. (i.e. `['HtmlTag', ['tag' => 'div', 'ignore' => [FieldsetElement::class], 'class' => 'form-element']]`)

src/FormDecorator/BaseDecorator.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Register attribute callbacks
+     *
+     * Override this method in order to register attribute callbacks in concrete classes.
+     */
+    protected function registerAttributeCallbacks(Attributes $attributes): void
+    {
+    }
```
nilmerg, line 70: Since I suggested to not require a constructor argument for decorators, this should be part of a trait instead. i.e. `DecoratorOptions` 😁

src/FormElement/BaseFormElement.php
```diff
@@ -387,4 +396,83 @@ protected function getValueOfNameAttribute()
...
+
+        return $this;
+    }
+
+    /**
+     * Get the decorators
+     *
+     * @return ?Decorator
+     */
+    public function getDecorator(string $name): ?Decorator
```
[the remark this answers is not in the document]
sukhwinder33445: I will remove it later, as I still have use cases for it in my local tests.

src/FormElement/BaseFormElement.php
```diff
@@ -387,4 +396,83 @@ protected function getValueOfNameAttribute()
...
+    }
+
+    /**
+     * Add the decorators
+     *
+     * @param iterable $decorators
+     *
+     * @return $this
+     */
+    public function addDecorators(iterable $decorators): self
```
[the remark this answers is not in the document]
sukhwinder33445: This can be used to add further decorators (e.g. conditionally) to an element after it has been created.

src/FormElement/FormElements.php
```diff
@@ -506,4 +613,25 @@ public function remove(ValidHtml $elementOrHtml)
...
+     * @return string Decorated element
+     */
+    public function renderDecorated(FormElement $element): string
+    {
+        return $element->getDecorators()->apply($element);
+    }
+
+    public function renderElement(ValidHtml $element): string
+    {
+        if ($element instanceof FormElement && $element->hasDecorators() && ! $this->hasDefaultElementDecorator()) {
```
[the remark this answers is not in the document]
sukhwinder33445: We have decided to prioritize the legacy decorator if both are present. If an element has explicit decorators (new), it will use them and ignore the legacy one. No exception is thrown.

src/FormElement/BaseFormElement.php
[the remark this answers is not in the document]
sukhwinder33445: Removed.

src/Form.php
```diff
@@ -46,6 +47,18 @@ class Form extends BaseHtmlElement
...
+    public function __construct()
+    {
+        if ($this->getDefaultElementDecorators() === null && ! $this->hasDefaultElementDecorator()) {
+            $this->setDefaultElementDecorators([
+                'Fieldset',
+                'Label',
+                'Description',
+                'HtmlTag' => ['tag' => 'div'],
+            ]);
+        }
```
nilmerg, line 59: Why is this not part of the class declaration already?

src/FormElement/FormElements.php
```diff
@@ -29,12 +30,53 @@ trait FormElements
     /** @var bool Whether the default element loader has been registered */
     protected $defaultElementLoaderRegistered = false;
 
+    /**
+     * Custom Element decorator loader paths
+     *
+     * @var array<int, array<string|int, string> Override this property to add custom decorator loader paths
+     *
+     * Each entry must be an array, where the key is the decorator name postfix (if any) and the value is the path
```
nilmerg, line 38: That's usually right after the first line.

src/FormElement/FormElements.php
```diff
@@ -157,6 +216,33 @@ public function createElement($type, $name, $options = null)
         /** @var FormElement $element */
         $element = new $class($name);
 
+        $customDecoratorPaths = $this->elementDecoratorLoaderPaths;
+        if (! empty($customDecoratorPaths)) {
+            if ($element instanceof DefaultFormElementDecoration) {
+                $element->addElementDecoratorLoaderPaths($customDecoratorPaths);
```
nilmerg, line 222: This misses the point of the interface check, since this method is not part of it.

Also, isn't this redundant anyway? The loop below passes the loaders to the element's chain already. Why is there a distinction between *default* loaders and *individual* ones?

src/FormDecorator/DecoratorChain.php
```diff
@@ -0,0 +1,223 @@
...
+    public function __construct()
+    {
+        $this->addDefaultPluginLoader('decorator', __NAMESPACE__, 'Decorator');
+    }
+
+    /**
+     * Add a decorator loader
+     *
+     * @param string $namespace Namespace of the decorators
+     * @param string $postfix   Decorator name postfix, if any
```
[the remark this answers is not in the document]
sukhwinder33445: I took this from `Plugins::addPluginLoader()` so that the names of the parameters match. I have changed it to Suffix for this method.

src/Form.php
sukhwinder33445: Tried, but unfortunately trait properties cannot be declared in the class.

src/Contract/DecoratorOptions.php
```diff
@@ -0,0 +1,52 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Register attribute callbacks
+     *
+     * Override this method in order to register attribute callbacks in concrete classes.
+     */
+    protected function registerAttributeCallbacks(Attributes $attributes): void
```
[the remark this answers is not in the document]
sukhwinder33445: The Decorator interface requires the getAttributes() method, which is defined in this trait. Every Decorator therefore uses this trait, but only a few need to define callbacks.

src/FormDecorator/HtmlTagDecorator.php
```diff
@@ -0,0 +1,150 @@
...
+
+    /** @var class-string<FormElement>[] Element name to ignore when decorating */
+    protected array $ignore = [];
+
+    /** @var string|string[] CSS classes to apply */
+    protected string|array $class = 'form-element';
+
+    public function decorate(DecorationResults $results, FormElement $formElement): void
+    {
+        if (in_array($formElement::class, $this->getIgnore(), true)) {
```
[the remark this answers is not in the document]
sukhwinder33445: I have added a `condition` (callable) attribute instead, as discussed offline.

src/Contract/DecoratorOptions.php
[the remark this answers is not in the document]
nilmerg: Unless I'm misunderstanding something, the only decorator not using callbacks is the fieldset decorator, but it doesn't use attributes as well. So which decorator are you speaking of?

src/FormDecorator/DecoratorChain.php
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * @param FormElement $formElement The form element to decorate
+     *
+     * @return string The decorated form element
+     */
+    public function apply(FormElement $formElement): string
+    {
+        $results = (new DecorationResults())->append($formElement);
+
+        if (! $formElement instanceof HiddenElement) {
```
nilmerg, line 248: We should reconsider how hidden elements are handled. The fact that they are ignored here is definitely wrong. The have decorators after all if it comes to this, so they should be decorated.

This brings me to the actual solution: Do not apply default decorators to hidden elements.

This way they won't be decorated *by default*. This does not exempt them from being decorated though, in case the are explicitly decorated.

src/FormElement/FormElements.php
```diff
@@ -157,6 +202,27 @@ public function createElement($type, $name, $options = null)
...
+            }
+
+            foreach ($customDecoratorPaths as $path) {
+                $elementDecoratorChain->addDecoratorLoader($path[0], $path[1] ?? '');
+            }
+        }
+
+        $defaultDecorators = $this->getDefaultElementDecorators();
+        if (! empty($defaultDecorators) && ! $this->hasDefaultElementDecorator()) {
```
nilmerg, line 218: Ignore hidden elements here. Please note that I think it makes sense to also ignore elements with the `hidden` attribute set to true. (Not just instances of `HiddenElement`)

nilmerg (changes requested):

src/FormDecorator/DescriptionDecorator.php
```diff
@@ -0,0 +1,80 @@
...
+     * @return $this
+     */
+    public function setClass(string|array $class): static
+    {
+        $this->class = $class;
+
+        return $this;
+    }
+
+
```
nilmerg, line 47: single empty line

src/FormDecorator/Placement.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace ipl\Html\FormDecorator;
+
+/**
+ * Describes the placement of decorating element
```
nilmerg, line 6: Rephrase, incorrect grammar

src/FormDecorator/HtmlTagDecorator.php
```diff
@@ -0,0 +1,177 @@
...
+use function ipl\Stdlib\get_php_type;
+
+/**
+ * Decorates the form element with an HTML tag
+ */
+class HtmlTagDecorator implements Decorator, DecoratorOptionsInterface
+{
+    use DecoratorOptions;
+
+    /** @var Placement Describes where the HTML tag should be placed. Default : Wrap */
```
nilmerg, line 22: why the space before the colon?

src/FormDecorator/HtmlTagDecorator.php
```diff
@@ -0,0 +1,177 @@
...
+    protected Placement $placement = Placement::Wrap;
+
+    /** @var string HTML tag to use for the decoration. */
+    protected string $tag;
+
+    /** @var ?callable(FormElement): bool Callable to decide whether to decorate element */
```
nilmerg, line 28: *decorate* **an/the** *element*

nilmerg (changes requested):

src/FormDecorator/HtmlTagDecorator.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    public function getCondition(): ?callable
+    {
+        return $this->condition;
+    }
+
+    /**
+     * Set the element tags to ignore when decorating
+     *
+     * @param callable(FormElement & HtmlElementInterface): bool $condition
```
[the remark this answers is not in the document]
nilmerg: That's not the case anymore with https://github.com/Icinga/ipl-html/pull/158, so you might need to re-introduce the intersection type.

PR #150, opened 2025-08-08
Escape special chars in form element name

### LINKED ISSUE #149: Form data and PHP $_POST data mismatch in case of special characters in form element names
In PHP, POST data automatically converts dots (`.`) and spaces (` `) in form element names to underscores (`_`). As a result, if form element names include these characters, the submitted data may not be correctly saved, leading to improper form population.

In Zend Framework, this issue is mitigated by stripping out special characters from element names. When it is not possible to exclude special characters, a name mapping strategy is implemented in the corresponding form to accurately persist and retrieve the associated data.

raviks789:
fixes #149


src/FormElement/BaseFormElement.php
```diff
@@ -137,6 +164,35 @@ public function setIgnored($ignored = true)
...
+     * The name is sanitized to only contain alphanumeric characters, underscores and brackets.
+     *
+     * @param  string $value
+     *
+     * @return string
+     */
+    public static function sanitizeName($value, $escapeBrackets = true)
+    {
+        $value = mb_strtolower($value ?? '', 'UTF-8');
+        $charset = '^a-zA-Z0-9_';
```
nilmerg, line 178: I think we should only replace reserved chars now. As per https://www.php.net/manual/en/language.variables.external.php, the only chars are: `.`, ` `, `[`, `]`

> Dots and spaces in variable names are converted to underscores. For example \<input name="a.b" /\> becomes $_REQUEST["a_b"].

> If an external variable name begins with a valid array syntax, trailing characters are silently ignored. For example, \<input name="foo[bar]baz"\> becomes $_REQUEST['foo']['bar'].

Please also add a comment in the code including this link.

src/FormElement/BaseFormElement.php
```diff
@@ -115,6 +118,30 @@ public function setName($name)
     {
         $this->name = $name;
 
+        // Name is always sanitized
+        if (! isset($this->sanitizedName)) {
```
nilmerg, line 122: Why this check? If a new name is set, is there a reason that the sanitized name must not change?

nilmerg (changes requested):

src/FormElement/FormElements.php
```diff
@@ -186,7 +191,24 @@ public function registerElement(FormElement $element)
...
+            $this->escapedElementNameCounter[$sanitizedName]++;
+            $sanitizedName .= $this->escapedElementNameCounter[$sanitizedName];
+            $element->setSanitizedName($sanitizedName);
+        }
+
         $this->elements[$name] = $element;
+        $this->escapedElementNameCounter[$sanitizedName] = 1;
+
+        // If the populated values is the POST data, use the sanitized name
+        if (array_key_exists($sanitizedName, $this->populatedValues)) {
```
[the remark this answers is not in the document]
raviks789: I am not sure about this, wouldn't that mean we would have to call `$this->getPopulatedValue($sanitizedName)` instead of `$this->getPopulatedValue($name)` to fetch the populated value. And moreover, the elements are not yet registered and we might not know the name->sanitizedName mapping yet.

Maybe we could also discuss this offline further.

src/FormElement/BaseFormElement.php
raviks789: Just for a confirmation, you mean other than the characters `.`, ` ` , `[`, `]`, other special characters does not need to be escaped right?

src/FormElement/FormElements.php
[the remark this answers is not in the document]
nilmerg: `getPopulatedValue(…)` might perform an automatic fallback if nothing is found, by sanitizing the given name and repeating the check

src/FormElement/BaseFormElement.php
nilmerg: Yes.

src/FormElement/FormElements.php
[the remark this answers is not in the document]
nilmerg: Let's just drop the counting altogether. If we don't replace all chars with an underscore, but with their own replacement, we preserve their meaning, have no duplicates and no problems detecting which sanitized name belongs to which element, without having an element.

So please find a safe character to use for each of the four.

src/Form.php
```diff
@@ -59,6 +59,40 @@ public static function isEmptyValue($value): bool
...
+     * @param string $str The string to escape
+     * @param bool   $escapeBrackets Whether to escape brackets
+     *
+     * @return string
+     */
+    public static function escapeReservedChars(string $str, bool $escapeBrackets = false): string
+    {
+        $escapeMap = [
+            '.' => chr(28), // File Separator
+            ' ' =>  chr(29) // Group Separator
```
nilmerg, line 85: superfluous space

src/Form.php
```diff
@@ -59,6 +59,40 @@ public static function isEmptyValue($value): bool
...
+     * And if an external variable name begins with a valid array syntax, trailing characters are silently ignored.
+     * For example, <input name="foo[bar]baz"> becomes $_REQUEST['foo']['bar'].
+     * See https://www.php.net/manual/en/language.variables.external.php
+     *
+     * @param string $str The string to escape
+     * @param bool   $escapeBrackets Whether to escape brackets
+     *
+     * @return string
+     */
+    public static function escapeReservedChars(string $str, bool $escapeBrackets = false): string
```
nilmerg, line 81: `$string`, please

nilmerg (changes requested):
Please also test the changes made to `populate()` and `registerElement()`. And those made to `BaseFormElement::setName()`.

src/Form.php
nilmerg, line 81: brackets should be escaped by default I guess.

src/FormElement/FormElements.php
```diff
@@ -249,6 +250,7 @@ public function createElement($type, $name, $options = null)
      */
     public function registerElement(FormElement $element)
     {
+        $escapedName = $element->getEscapedName();
```
nilmerg, line 253: This requires an interface check as well!

nilmerg (changes requested):

nilmerg:
And please fix the conflicts

PR #154, opened 2025-08-25
Introduce `HtmlElementInterface`

sukhwinder33445:


src/Contract/HtmlElementInterface.php
```diff
@@ -0,0 +1,87 @@
...
+     * Get the attributes of the element
+     *
+     * @return Attributes
+     */
+    public function getAttributes(): Attributes;
+
+    /**
+     * Add the given attributes to the element
+     *
+     * @param Attributes|array $attributes
```
nilmerg, line 31: Please add a phpstan-type to `Attributes` decribing that `array` structure and import and use it here.

src/Contract/HtmlElementInterface.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public function getAttribute(string $name): Attribute;
+
+    /**
+     * Set the attribute with the given name and value
+     *
+     * If the attribute with the given name already exists, it gets overridden.
+     *
+     * @param string $name  The name of the attribute
+     * @param bool|string|array $value The value of the attribute
```
nilmerg, line 65: Same here, please describe this in `Attribute` and import the type here.

nilmerg (changes requested):

src/Attribute.php
```diff
@@ -12,6 +12,8 @@
  *
  * Usually attributes are not instantiated directly, but created through an HTML
  * element's exposed methods.
+ *
+ * @phpstan-type AttributeValueFormat string|bool|string[]|null
```
nilmerg, line 16: Why the `Format` suffix? It's a type after all.

```suggestion
 * @phpstan-type _AttributeScalar string|bool|null
 * @phpstan-type AttributeValue _AttributeScalar|array<_AttributeScalar>
```

src/Attribute.php
```diff
@@ -91,7 +93,7 @@ public static function escapeName($name)
      * Values are escaped according to the HTML5 double-quoted attribute value syntax:
      * {@link https://html.spec.whatwg.org/multipage/syntax.html#attributes-2 }.
      *
-     * @param string|array $value
+     * @param string|string[] $value
```
nilmerg, line 96: Why does this not use the phpstan type? phpstan thinks this too.

src/Attribute.php
nilmerg, line 1: Please clean the phpstan baseline up then, there are no errors left after this.

src/Attributes.php
```diff
@@ -17,6 +17,10 @@
  * behavior in various ways.
  *
  * Attributes usually come in name-value pairs and are rendered as name="value".
+ *
+ * @phpstan-import-type AttributeValueFormat from Attribute
+ *
+ * @phpstan-type AttributesFormat array<int|string, Attribute|AttributeValueFormat>
```
nilmerg, line 23: This cannot be correct. This is:

`array<string, AttributeValue>`

Passing `[new Attribute('foo', 'bar')]` to `Attributes::add(…)` will register the attribute with the name `0` and the value set to the attribute object.

nilmerg: And again, do not use `Format` as suffix. Use `Type` if necessary.

nilmerg (changes requested):

nilmerg (approved):

PR #155, opened 2025-08-26
Add property and method return types

sukhwinder33445:
Dependencies:
- https://github.com/Icinga/ipl-web/pull/312
- https://github.com/Icinga/gipfl-web/pull/2


nilmerg:
But let's keep the branch for now, it will get its chance

PR #156, opened 2025-08-27
Deprecate interface Wrappable

sukhwinder33445:
The implementation of this interface in the `HtmlDocument` class requires the property of the class (`HtmlDocument::renderWrapped()` uses the `renderedBy` property). This means that a `Wrappable` instance must actually also be an `HtmlDocument` instance. This in turn means that `Wrappable` can be removed (deprecate) and the `HtmlDocument` can be used directly instead.

### DIFF
--- a/src/HtmlDocument.php
+++ b/src/HtmlDocument.php
@@ -28,14 +28,14 @@ class HtmlDocument implements Countable, Wrappable, MutableHtml
     /** @var bool Whether the document has been assembled */
     protected $hasBeenAssembled = false;
 
-    /** @var Wrappable Wrapper */
-    protected $wrapper;
+    /** @var ?(HtmlDocument|Wrappable) Wrapper */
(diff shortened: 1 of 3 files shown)


nilmerg (changes requested):

src/HtmlDocument.php
```diff
@@ -379,7 +379,10 @@ protected function renderWrapped()
         $wrapper = $this->wrapper;
 
         if (isset($this->renderedBy)) {
-            if ($wrapper === $this->renderedBy || $wrapper->contains($this->renderedBy)) {
+            if (
+                $wrapper === $this->renderedBy
+                || ($wrapper instanceof MutableHtml && $wrapper->contains($this->renderedBy))
+            ) {
```
nilmerg, line 385: This has nothing to do with the deprecation. Please open a separate PR.

nilmerg (changes requested):

nilmerg (approved):

PR #157, opened 2025-09-08
Support `ipl\Stdlib\Option` for form elements

nilmerg:


nilmerg:
Just noticed, annotating properties is not an *option* (pun intended) because of inheritance when previously a callback was registered. A subclass may override the setter, which is bypassed when annotating only the property.

cc @lippserd

PR #158, opened 2025-09-09
Revert all latest changes regarding type declarations

nilmerg:


src/HtmlDocument.php
```diff
@@ -201,12 +201,12 @@ public function addHtml(ValidHtml ...$content): static
...
      *
-     * @param MutableHtml $from
-     * @param callable    $callback Optional callback in order to transform the content to add
+     * @param HtmlDocument $from
+     * @param callable     $callback Optional callback in order to transform the content to add
      *
      * @return $this
      */
-    public function addFrom(MutableHtml $from, $callback = null)
+    public function addFrom(HtmlDocument $from, $callback = null)
```
sukhwinder33445, line 209: Why not `MutableHtml` ?

src/Contract/FormElement.php
```diff
@@ -2,13 +2,30 @@
...
+     * Get the attributes or options of the element
+     *
+     * @return Attributes
+     */
+    public function getAttributes();
+
+    /**
+     * Add attributes or options to the form element
+     *
+     * @param iterable $attributes
```
sukhwinder33445, line 23: Please use `phpstan-type` defined in the `Attributes` class.
`HtmlElementInterface` already uses it.

src/HtmlDocument.php
nilmerg: image

sukhwinder33445: I see 😞 .

sukhwinder33445 (approved):

PR #161, opened 2025-09-15
RadioElement: Provide a tag in order to be decorated

nilmerg:
The reason why #160 exists.


src/FormElement/RadioElement.php
```diff
@@ -24,6 +24,14 @@ class RadioElement extends BaseFormElement
     /** @var array Disabled radio options */
     protected $disabledOptions = [];
 
+    protected function tag()
```
sukhwinder33445, line 27: Please add return type.

sukhwinder33445 (approved):

PR #163, opened 2025-09-17
Apply decorators not too late

nilmerg:
The new decorators are applied too late. It's not possible to e.g. change elements depending on others nor to prepend or append additional content to their form. When `HtmlDocument::renderElement()` runs, the form's content has already been finalized with the call to `::getContent()` inside `::renderUnwrapped()`.

To change this, decorators are now applied after form assembly. Application of decorators is now also done differently as they're registered as wrappers again. But this time not in a way that anyone can or should assume their existence. Once a form has been assembled, the decoration results are *added* as wrapper. This means that any other wrappers the elements already have, will be decorated as well.

* Elements will always be rendered, even without a decorator that assigns them a position (default behavior of wrappers)
* To decorate an element, **`FormElements::decorate()` must be used**, but that should not be a problem as it already used to be like this.
  At least for *new* default decorators. *Legacy* custom decorators didn't require this, but anyone who's not using `addElement` always had to replicate it's behavior so this is fine to me

requires #162

### DIFF
--- a/src/FormDecoration/DecorationResults.php
+++ b/src/FormDecoration/DecorationResults.php
@@ -11,7 +11,7 @@ use ipl\Html\ValidHtml;
  *
  * @phpstan-type content array<int, ValidHtml|array<int, mixed>>
  */
-class DecorationResults implements ValidHtml
+class DecorationResults
 {
     /** @var content The HTML content */
     protected array $content = [];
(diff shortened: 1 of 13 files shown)


sukhwinder33445 (approved):
👍🏽 🥳

PR #167, opened 2025-09-19
Enhance label and description decorator

sukhwinder33445:
Blocked by: https://github.com/Icinga/ipl-html/pull/168

### DIFF
--- a/src/FormDecoration/LabelDecorator.php
+++ b/src/FormDecoration/LabelDecorator.php
@@ -63,8 +63,13 @@ class LabelDecorator implements FormElementDecoration, DecoratorOptionsInterface
 
         if ($elementLabel instanceof HtmlElementInterface) {
             $attributes['class'] = $this->getClass();
-            if ($isHtmlElement && $formElement->getAttributes()->has('id')) {
-                $attributes['for'] = $formElement->getAttributes()->get('id')->getValue();
+            if ($isHtmlElement) {
+                $elementAttributes = $formElement->getAttributes();
+                if (! $elementAttributes->has('id')) {
+                    $elementAttributes->set('id', uniqid('form-element-'));
+                }
+
+                $attributes['for'] = $elementAttributes->get('id')->getValue();
             }
 
             $elementLabel->addAttributes($attributes);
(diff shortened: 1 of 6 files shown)


src/FormDecoration/LabelDecorator.php
```diff
@@ -55,20 +56,47 @@ public function decorate(DecorationResults $results, FormElement $formElement):
...
+     */
+    protected function getElementLabel(FormElement $formElement): ?ValidHtml
+    {
+        $label = $formElement->getLabel();
+        if ($label === null) {
+            return null;
         }
 
-        $results->append(new HtmlElement('label', $labelAttr, new Text($formElement->getLabel())));
+        return new HtmlElement('label', content: new Text($label));
```
nilmerg, line 99: It's the label decorator, so I think this should not be possible to influence by child classes. only return text here.

nilmerg (changes requested):

src/FormDecoration/LabelDecorator.php
```diff
@@ -47,24 +51,67 @@ public function setClass(string|array $class): static
...
     }
 
+    /**
+     * Set whether to apply the fallback id to the element if not set
+     *
+     * @param bool $apply
+     *
+     * @return $this
+     */
+    public function applyFallbackId(bool $apply = true): static
```
nilmerg, line 61: Register this as attribute callback please

tests/FormDecorator/LabelDecoratorTest.php
```diff
@@ -37,15 +37,49 @@ public function testSetClass(): void
...
+
+    /**
+     * @depends testMethodApplyFallbackId
+     */
     public function testWithLabelAttributeOnly(): void
     {
         $results = new FormElementDecorationResult();
-        $this->decorator->decorateFormElement($results, new TextElement('test', ['label' => 'Label Here']));
+        $this->decorator
+            ->applyFallbackId(false) // disable fallback ID
```
nilmerg, line 79: You've given the element an id in the other test, why not here as well? Saves you the test dependency at least.

nilmerg (changes requested):

tests/FormDecorator/LabelDecoratorTest.php
sukhwinder33445: Yes, but then it doesn't fit with the name of the test.

nilmerg (changes requested):

nilmerg (approved):

PR #170, opened 2025-09-26
Deprecate everything related to the old form element decorators

nilmerg:


nilmerg:
> Method FormElements::hasDecorator() can also be deprecated.

There's no such thing.

PR #176, opened 2025-10-17
DecoratorChain: Allow to replace specific decorators

nilmerg:
For this all decorators require an identifier now. This is a breaking change compared to the current state. In the past using Zend form, we often copied the default list of decorators just to change a single decorator. Though, changes to the default decorators were never applied by forms overriding them, so this functionality in the new implementation should provide more flexibility in this regard.

### DIFF
--- a/src/FormDecoration/DecoratorChain.php
+++ b/src/FormDecoration/DecoratorChain.php
@@ -132,45 +132,50 @@ class DecoratorChain implements IteratorAggregate
      *
      * The order of the decorators is important, as it determines the rendering order.
      *
+     * > NOTE: Decorators are registered with unique identifiers to enable easy replacement. If the decorator is
+     * provided as a string and no identifier is given, the decorator name is used as a fallback. In all other cases,
+     * an identifier must be specified explicitly.
+     *
+     *  *It is recommended to use the decorator name or its purpose as the identifier, for example, 'container' for an
+     * 'HtmlTag' decorator that wraps content.*
+     *
      *   *The following array formats are supported:*
      *
      * ```
-     * // When no options are required or defaults are sufficient
+     * // (1) When no options are required or defaults are sufficient
      * $decorators = [
      *     'Label',
      *     'Description'
      * ];
      *
-     * // Custom identifiers
-     * $decorators = [
-     *     'label' => 'Label',
-     *     'description' => 'Description'
-     * ];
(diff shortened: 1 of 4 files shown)


src/FormDecoration/DecoratorChain.php
```diff
@@ -132,45 +132,53 @@ public function replaceDecorator(string $identifier, object|string $decorator, a
...
-     *     ]
+     *          'name'      => 'HtmlTag',
+     *          'options'   => ['tag' => 'div']
+     *      ]
      * ];
      *
-     * // or add Decorator instances
+     * // (4) Add Decorator instances
+     * $htmlTag = new HtmlTagDecorator();
+     * $htmlTag->getAttributes()->add(['tag' => 'div']);
```
nilmerg: This is still wrong. `HtmlTagDecorator`'s attributes are not meant to be used like this, there's

a) a setter for the tag
b) arbitrary attributes need to be defined with `setAttrs()`

src/FormDecoration/DecoratorChain.php
```diff
@@ -200,15 +200,19 @@ public function addDecorators(DecoratorChain|array $decorators): static
             if (is_int($identifier)) {
                 if (! is_string($decoratorOptions)) {
                     throw new InvalidArgumentException(sprintf(
-                        'Unexpected type at position %d, string expected, got %s instead',
+                        'Unexpected type at position %d, string expected, got %s instead.'
+                        . ' Use an identifier (string) instead of key %d',
                         $identifier,
-                        get_php_type($decoratorOptions)
+                        get_php_type($decoratorOptions),
+                        $identifier
```
nilmerg: redundant, as discussed, the same applies in the message below

tests/FormDecorator/DecoratorChainTest.php
```diff
@@ -291,4 +313,67 @@ public function testReplaceDecoratorPreservesDecoratorOrder(): void
...
+                "Invalid type for identifier 'test', expected an array, a string or an instance of %s,"
+                ." got integer instead",
+                FormElementDecoration::class
+            )
+        );
+
+        $this->chain->addDecorators(['test' => 5]);
+    }
+
+    public function testAddDecoratorWithDifferentTypeThanTheChainAcceptsThrowsException(): void
```
nilmerg: please get rid of that, as discussed

sukhwinder33445 (approved):

PR #177, opened 2025-10-28
FieldsetElement: Remove superfluous `aria-describedby` attribute

sukhwinder33445:


src/FormDecoration/FieldsetDecorator.php
```diff
@@ -25,14 +25,7 @@ public function decorateFormElement(DecorationResult $result, FormElement $formE
 
         $description = $formElement->getDescription();
         if ($description !== null) {
```
nilmerg: and please add a comment, why this hasn't any id fallback handling or aria-describedby

nilmerg (changes requested):

nilmerg (approved):

PR #183, opened 2025-11-27
BaseFormElement: Accept null as name in `setName`

### LINKED ISSUE #182: BC in v0.9.0 due to form elements not accepting `null` as name
#150 introduced a BC, knowingly. Though, no one was aware of it being a real issue: https://github.com/Icinga/icingaweb2/issues/5453

nilmerg:
fixes #182
closes https://github.com/Icinga/icingaweb2/issues/5453


raviks789 (approved):
LGTM!

PR #184, opened 2025-12-03
Support PHP 8.5

### LINKED ISSUE #181: PHP 8.4/8.5 Support

Jan-Schuppik:
### PHP 8.4:
- [Function parameters that are null by default must be declared nullable](https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core).

### PHP 8.5:
- [Using null as an array offset is deprecated.](https://www.php.net/manual/en/migration85.deprecated.php#migration85.deprecated.core.using-null-as-an-array-offset)
- Remove deprecated method [`ReflectionProperty::setAccesible()`](https://www.php.net/manual/en/migration85.deprecated.php#migration85.deprecated.reflection)

### Other changes:
- Update composer package `guzzlehttp/psr7`

resolves #181

### DIFF
--- a/src/FormElement/RadioElement.php
+++ b/src/FormElement/RadioElement.php
@@ -66,7 +66,7 @@ class RadioElement extends BaseFormElement
      *
      * @throws InvalidArgumentException If no option with the specified value exists
      */
-    public function getOption($value): RadioOption
+    public function getOption(string|int $value): RadioOption
     {
         if (! isset($this->options[$value])) {
             throw new InvalidArgumentException(sprintf('There is no such option "%s"', $value));
@@ -127,10 +127,15 @@ class RadioElement extends BaseFormElement
                     'checked',
                     function () use ($option) {
                         $optionValue = $option->getValue();
+                        $value = $this->getValue();
+
+                        if ($optionValue === '' && $value === null) {
+                            return true;
+                        }
 
                         return ! is_int($optionValue)
-                            ? $this->getValue() === $optionValue
-                            : $this->getValue() == $optionValue;
+                            ? $value === $optionValue
+                            : $value == $optionValue;
                     }
                 )
                 ->registerAttributeCallback(
--- a/tests/FormElement/RadioElementTest.php
+++ b/tests/FormElement/RadioElementTest.php
@@ -282,7 +282,7 @@ HTML;
         $this->assertFalse($radio->isValid());
     }
 
-    public function testNullAndTheEmptyStringAreEquallyHandled()
+    public function testNullAndTheEmptyStringValuesAreEquallyHandled()
     {
         $form = new Form();
         $form->addElement('radio', 'radio', [
@@ -303,8 +303,6 @@ HTML;
         $this->assertNull($radio2->getValue());
 
         $this->assertInstanceOf(RadioOption::class, $radio->getOption(''));
-        $this->assertInstanceOf(RadioOption::class, $radio2->getOption(null));
-        $this->assertInstanceOf(RadioOption::class, $radio->getOption(null));
         $this->assertInstanceOf(RadioOption::class, $radio2->getOption(''));
 
         $this->assertTrue($radio->isValid());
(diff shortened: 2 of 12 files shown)


sukhwinder33445 (dismissed):
Looks good

sukhwinder33445 (dismissed):

BastianLedererIcinga (dismissed):

sukhwinder33445 (dismissed):

src/FormElement/RadioElement.php
```diff
@@ -68,6 +68,9 @@ public function setOptions(array $options): self
      */
     public function getOption($value): RadioOption
     {
+        // php>=8.5 does not support null as array key
+        $value = $value ?? '';
+
```
lippserd: The following question applies to all other code locations that you have introduced in this way: Why is this necessary? Because PHPDoc does not allow `null`.

tests/FormElement/RadioElementTest.php
```diff
@@ -282,6 +283,7 @@ public function testRadioNotValidIfCheckedValueIsDisabled()
         $this->assertFalse($radio->isValid());
     }
 
+    #[RequiresPhp('< 8.5')]
```
lippserd: Since using `null` as an array offset is deprecated and should be discouraged, I wouldn't even show anymore that we support this for PHP < 8.5. I suggest removing all test cases where `null` is used as an array offset. Please pay special attention to tests where the empty string and `null` are tested differently, e.g., for `value`.

lippserd (changes requested):
In addition to the requested changes, I have a few comments on the commit messages and descriptions:

For 982ef2846935678e133983d82059114cf827bffe, please use something like the following, as suggested in https://github.com/Icinga/ipl-stdlib/pull/61#pullrequestreview-3592029375:

```
PHP 8.4: Change implicit nullable type declaration to explicit

Since PHP 8.4 implicitly nullable parameter types are deprecated.
```

The reason for upgrading `guzzlehttp/psr7` in d5b09891733a7004724259c1c3864dfc1d34f789 is misleading, as version 2.8.0 introduces PHP 8.5 support and is a dependency for this library. In general, our ipl-* repositories build our Icinga PHP library, not the other way around. Therefore, we should never have such "streamline" commits. Something similar was also discussed/addressed in https://github.com/Icinga/ipl-validator/pull/38.

For 41413f04793a8c9fc9d7a2bd65bc64e2ec678b11, the commit message could be worded more precisely, and a description justifying the removal should be added, e.g.:

```
PHP 8.5: Remove deprecated method `ReflectionProperty::setAccessible()`

Since we require PHP >= 8.2 and since calling this method has no effect as of
PHP 8.1.0, as all properties are accessible by default, it is safe to simply
remove this method call.
```

Finally, 3ef2e009b2879316d6effb2b78a2d88d5d95bdeb could be rephrased to ```PHP 8.5: Replace usage of `null` as array offset```. A description that only contains a reference to the PHP documentation is not necessary, but something like the following:

```
Since PHP 8.5 using `null` as an array offset is deprecated. Code relying on 
`null` being silently converted to an empty string must be replaced and
discouraged in PHPDocs.
```

lippserd:
And please also add a PR description.

lippserd (changes requested):
Both concerns regarding the radio element also apply to the select element.

src/FormElement/RadioElement.php
```diff
@@ -44,10 +44,7 @@ public function setOptions(array $options): self
         $this->options = [];
         foreach ($options as $value => $label) {
             $option = (new RadioOption($value, $label))
-                ->setDisabled(
-                    in_array($value, $this->disabledOptions, ! is_int($value))
-                    || ($value === '' && in_array(null, $this->disabledOptions, true))
-                );
+                ->setDisabled(in_array($value, $this->disabledOptions));
```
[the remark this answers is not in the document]
lippserd: I'm sorry, I misread the code. Other types were also applied previously, only the empty string was handled specially. Without your changes, the code would result in the following:

```php
var_dump($radio->getOption('')->isDisabled()); // bool(false)
var_dump($radio->getOption(0)->isDisabled()); // bool(true)
var_dump($radio->getOption(1)->isDisabled()); // bool(true)
```

Of course, it still makes sense to fix this. And I would introduce tests for the other types so that we don't accidentally change the behavior.

lippserd (changes requested):
Since explicit is better than implicit, I also suggest adding new `assertNull()` calls after the `setValue()` calls at the end of `testNullAndTheEmptyStringValuesAreEquallyHandled()` for both `Radio` and `SelectElementTest`. If I'm not mistaken, there is no check to validate if `setValue(''|null)` leads to `getValue(): null`.

lippserd (approved):
Thanks 👍

PR #189, opened 2026-03-16
Adapt strict type declarations in `FormElements::isValidEvent()`

BastianLedererIcinga:
Add strict types to `FormElements::isValidEvent()`.

requires https://github.com/Icinga/ipl-stdlib/pull/69


lippserd (commented):
That's also a forward-compatible change, right? If so, remove the `composer.json` bump and update the PR title and description accordingly.

PR #191, opened 2026-03-19
Add `Form::yieldElementsRecursive()` to fix partial validation for `FormElements`

### LINKED ISSUE #190: Required fields of fieldsets are invalid on autosubmit
### Description

One example is the `ChannelForm` of Icinga Notifications Web. When changing the `Type` dropdown from another value to 'Email', the following happens:

Image

### Cause

When a form is submitted via autosubmit, it is only validated partially. This means each element is checked individually to determine whether it has a value, and only validate it if so. For fieldset elements, however, it is sufficient for one child element to have a value for the entire fieldset to be considered as having a value. As a result, `$element->validate()` is called on the fieldset, which fully validates every child element, including ones that have just appeared and are still empty.

### Solution

We need to check whether an element is a fieldset. If so, call a new `validatePartial()` method on it instead, which checks each child element individually for whether it has a value before validating it fully.

`FieldsetElement::validatePartial()` should also handle the case where a child element itself is a fieldset.

jrauh01:
When a form is submitted via autosubmit, only partial validation runs. For elements containing other form elements (`FormElements` implementation), a single child having a value was enough to trigger full validation of the entire element, causing "field is required errors" to appear for fields that had just been revealed and were still empty.

Because initiating validation is the concern of `Form`, a `yieldElementsRecursive()` method is added there. It recursively yields all sub form elements and recurses into nested `FormElements` containers. This enables the form to check each individual element and validate only if it has a value.

fixes #190

### DIFF
--- a/src/Contract/FormElements.php
+++ b/src/Contract/FormElements.php
@@ -100,4 +100,11 @@ interface FormElements extends EventEmitterInterface
      * @return $this
      */
     public function populate(iterable $values);
+
+    /**
+     * Validate all elements that have a value
+     *
+     * @return $this
+     */
+    public function validatePartial(): static;
 }
--- a/src/Contract/FormElements.php
+++ b/src/Contract/FormElements.php
@@ -3,6 +3,7 @@
 namespace ipl\Html\Contract;
 
 use Evenement\EventEmitterInterface;
+use Generator;
 use InvalidArgumentException;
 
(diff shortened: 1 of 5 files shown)


nilmerg (commented):
Otherwise it looks fine I think.

lippserd (approved):

PR #195, opened 2026-03-25
Preserve `checked` state of `disabled` checkbox on submit

sukhwinder33445:
Browsers do not submit `disabled` inputs. Since `CheckboxElement`'s `hidden` fallback element always used the `unchecked` value, a `checked` but `disabled` checkbox would appear `unchecked` on the next submit.
Fix `renderUnwrapped()` to use the `checked` value in the `hidden` element when the checkbox is `checked` and `disabled`, preserving its state across submissions.

fixes #192

### DIFF
--- a/src/FormElement/CheckboxElement.php
+++ b/src/FormElement/CheckboxElement.php
@@ -119,7 +119,11 @@ class CheckboxElement extends InputElement
     {
         $html = parent::renderUnwrapped();
 
-        return (new HiddenElement($this->getValueOfNameAttribute(), ['value' => $this->getUncheckedValue()])) . $html;
+        $value = $this->getAttribute('disabled')->getValue() && $this->isChecked()
+            ? $this->getCheckedValue()
+            : $this->getUncheckedValue();
+
+        return (new HiddenElement($this->getValueOfNameAttribute(), ['value' => $value])) . $html;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


TheSyscall (approved):
Works as expected.

Changing the value of the hidden element seems like it could have unintended consequences, but I can't think of any case where it wouldn't work.

sukhwinder33445:
The value of hidden element is only changed, when the checkbox is `checked` and `disabled`. Otherwise it still uses the `unchecked` value as before.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
  | State                   | Hidden value | Checkbox submitted | Server receives | Before | After |                                                                                                                                                                                                                                                                                                            
  |-------------------------|--------------|--------------------|-----------------|------------|-----------|                                                                                                                                                                                                                                                                                                            
  | Enabled + unchecked     | `n`          | no                 | `n`             | ✅   | ✅  |                                                                                                                                                                                                                                                                                                            
  | Enabled + checked       | `n`          | yes → `y` (wins)  | `y`             | ✅   | ✅  |                                                                                                                                                                                                                                                                                                             
  | Disabled + unchecked    | `n`          | no                 | `n`             | ✅   | ✅  |                                                                                                                                                                                                                                                                                                            
  | Disabled + checked      | `n` (now → **`y`**)| no                 | **`y`**         |  ❌  | ✅  |

PR #196, opened 2026-04-16
Revert form element escaping

nilmerg:
It turned out to be rather problematic to implicitly escape every form element name as we cannot guarantee that element names are not programmatically constructed already.

Examples:

`\ipl\Html\FormElement\CheckboxElement` renders a hidden element representing its unchecked state and uses `getValueOfNameAttribute` to define the name for it. This is done to support fieldsets containing a checkbox element. Though, fieldsets require special element names and hence this results in an constructed name with brackets being used as literal and hence the hidden element escapes it.

Another case is when an element with an escaped name is also capable of auto-submitting the form. Icinga Web will transmit the name of such an element in a special http request header value and since we used invisible control characters to escape form element names, this resulted in invalid header values.

--

It's for these reasons that we decided to drop support for implicit form element name escaping again and require form implementors to safe-guard themselves by explicitly escaping their names. `Form::escapeReservedChars()` has been deprecated and triggers a deprecation notice if used from now on.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -66,6 +66,26 @@ class Form extends BaseHtmlElement implements Contract\Form, Contract\FormElemen
         return $value === null || $value === [] || (is_string($value) && trim($value) === '');
     }
 
+    /**
+     * Return the given string unchanged.
+     *
+     * @param string $string
+     * @param bool $escapeBrackets
+     *
+     * @return string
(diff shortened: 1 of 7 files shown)


raviks789 (dismissed):

nilmerg:
`__CLASS__` was superfluous in the deprecation notice.

raviks789 (approved):

PR #202, opened 2026-07-17
Labels without escaping

jrauh01:
`LabelDecorator` and `FieldsetDecorator` previously wrapped every label in `Text`, which HTML-escapes its content on render. Passing a `ValidHtml` label therefore caused its markup to be double-escaped, and passing any `ValidHtml` that does not implement `__toString()` caused a fatal error during the string coercion in `Text::setContent()`.

Both decorators now detect `ValidHtml` labels and use them directly, wrapping only plain strings in `Text`. `FieldsetDecorator` requires its own fix because fieldset elements bypass `LabelDecorator` entirely and render the legend themselves.

The `$label` property, `setLabel()`, and `getLabel()` type annotations are updated to reflect that labels may be either a plain string or a `ValidHtml` object.

Tests cover string escaping, `HtmlString` pass-through, and a render-only `ValidHtml` implementation (no `__toString()`) for both decorators.

### DIFF
--- a/src/FormDecoration/LabelDecorator.php
+++ b/src/FormDecoration/LabelDecorator.php
@@ -98,7 +98,7 @@ class LabelDecorator implements FormElementDecoration, DecoratorOptionsInterface
             return null;
         }
 
-        return new HtmlElement('label', content: new Text($label));
+        return new HtmlElement('label', content: $label instanceof ValidHtml ? $label : new Text($label));
     }
 
     protected function registerAttributeCallbacks(Attributes $attributes): void
--- a/src/FormElement/BaseFormElement.php
+++ b/src/FormElement/BaseFormElement.php
@@ -12,6 +12,7 @@ use ipl\Html\Contract\ValueCandidates;
 use ipl\Html\Form;
 use ipl\Html\FormDecoration\DecoratorChain;
 use ipl\Html\FormDecoration\FormElementDecorationResult;
+use ipl\Html\ValidHtml;
 use ipl\I18n\Translation;
 use ipl\Stdlib\Messages;
 use ipl\Validator\ValidatorChain;
@@ -30,7 +31,7 @@ abstract class BaseFormElement extends BaseHtmlElement implements FormElement, V
     /** @var string Description of the element */
     protected $description;
 
-    /** @var string Label of the element */
+    /** @var string|ValidHtml|null Label of the element */
     protected $label;
 
     /** @var string Name of the element */
@@ -100,7 +101,7 @@ abstract class BaseFormElement extends BaseHtmlElement implements FormElement, V
     /**
      * Set the label of the element
      *
-     * @param string $label
+     * @param string|ValidHtml $label
      *
      * @return $this
      */
--- a/tests/FormDecorator/LabelDecoratorTest.php
+++ b/tests/FormDecorator/LabelDecoratorTest.php
@@ -5,11 +5,13 @@ namespace ipl\Tests\Html\FormDecorator;
 use ipl\Html\Contract\FormElement;
 use ipl\Html\FormDecoration\FormElementDecorationResult;
 use ipl\Html\FormDecoration\LabelDecorator;
+use ipl\Html\HtmlString;
 use ipl\Html\FormElement\FieldsetElement;
 use ipl\Html\FormElement\SubmitButtonElement;
 use ipl\Html\FormElement\SubmitElement;
 use ipl\Html\FormElement\TextElement;
 use ipl\Html\Test\TestCase;
+use ipl\Html\ValidHtml;
 
 class LabelDecoratorTest extends TestCase
 {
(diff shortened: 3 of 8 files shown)


src/FormDecoration/LabelDecorator.php
```diff
@@ -98,12 +126,13 @@ protected function getElementLabel(FormElement $formElement): ?ValidHtml
             return null;
         }
 
-        return new HtmlElement('label', content: new Text($label));
+        return new HtmlElement('label', content: $this->getEscaped() ? new HtmlString($label) : new Text($label));
```
TheSyscall: `ipl\Html\Text` can already do that.

```suggestion
        return new HtmlElement('label', content: (new Text($label))->setEscaped($this->escaped));
```

TheSyscall (changes requested):

src/FormDecoration/LabelDecorator.php
```diff
@@ -12,6 +12,7 @@
 use ipl\Html\Contract\HtmlElementInterface;
 use ipl\Html\FormElement\RadioElement;
 use ipl\Html\HtmlElement;
+use ipl\Html\HtmlString;
```
lippserd: Please remove this unused import. The final implementation checks `ValidHtml` and no longer references `HtmlString` in production code.

lippserd (changes requested):
Please address the following before merging.

Please update the public label contract to include `ValidHtml`. `LabelDecorator` and the new test accept it, but `FormElement::getLabel()` and `BaseFormElement::$label`/`setLabel()` still advertise strings, so PHPStan considers the new branch impossible and statically checked callers cannot use the feature. Document that strings are escaped while `ValidHtml` is trusted markup.

Please apply the same `ValidHtml` pass-through to fieldset legends. Fieldsets bypass `LabelDecorator`, and `FieldsetDecorator` still wraps every label with `Text::create()`; browser verification showed `HtmlString` markup being escaped, while a render-only `ValidHtml` implementation fails during string conversion. Add coverage for string, `HtmlString`, and render-only `ValidHtml` labels.

Please update the PR description to match the final design. It still promises the removed `escaped` option and accessors, while the current diff uses automatic `ValidHtml` detection.

src/FormDecoration/LabelDecorator.php
jrauh01: Done

jrauh01:
I extended `FieldsetDecorator` with the same `ValidHtml` pass-through that `LabelDecorator` already had, updated the public label contract to include `ValidHtml`, and added tests for string escaping, `HtmlString` pass-through, and a render-only `ValidHtml` implementation in `FieldsetDecoratorTest`.

src/FormElement/BaseFormElement.php
```diff
@@ -30,7 +31,7 @@ abstract class BaseFormElement extends BaseHtmlElement implements FormElement, V
     /** @var string Description of the element */
     protected $description;
 
-    /** @var string Label of the element */
+    /** @var string|ValidHtml Label of the element */
```
lippserd: The property has no default value, so PHP initializes it to `null`. `getLabel()` returns `null` when no label has been set. The interface docblock correctly reads `@return string|ValidHtml|null`. The missing `|null` creates an incorrect type signal for static analysis and IDE tooling:

```suggestion
    /** @var string|ValidHtml|null Label of the element */
```

tests/FormDecorator/LabelDecoratorTest.php
```diff
@@ -111,6 +113,51 @@ public function testThatSubmitElementsAndFieldsetElementsAreIgnored(): void
...
+            '<label for="test-id" class="form-element-label"><em>italic</em></label>',
+            $results->assemble()
+        );
+    }
+
+    public function testRenderOnlyValidHtmlLabelIsRenderedAsIs(): void
+    {
+        $results = new FormElementDecorationResult();
+        $element = new TextElement('test', ['id' => 'test-id']);
+        $element->setLabel(HtmlString::create('<em>italic</em>'));
```
lippserd: This call reads as a copy-paste artifact from the adjacent `testValidHtmlLabelIsRenderedAsIs` test:

```suggestion
```

lippserd (changes requested):
In addition to the requested changes:

`SubmitElement::setLabel()` with no `@param` annotation inherits the widened ValidHtml contract but its pipeline is string-only. A caller who passes a `ValidHtml` object based on the advertised interface will silently break form submission detection: `hasBeenPressed()` compares the stored object against a string with `===`, which is always false for objects. Add a narrowing `@param string $label` annotation to `SubmitElement::setLabel()`. A brief docblock sentence explaining that the label populates a value attribute and must therefore be a plain string is sufficient.

I am not fully aware of the deprecated `DivDecorator` usage but its fieldset path throws `TypeError` for non-Stringable `ValidHtml`. I would apply the same guard used in `FieldsetDecorator`: replace `Text::create($label)` with `$label instanceof ValidHtml ? $label : new Text($label)`.

jrauh01:
Narrowed the type via `@param` annotation for `SubmitElement::setLabel()` and added the conditional logic to render `ValidHtml` for fieldset legends via `DivDecorator`.

lippserd (approved):
Please clean up the commit history.

jrauh01:
Commit history is clean now.

PR #203, opened 2026-07-17
Replace error list with individual `Callout` widgets in `onError`

jrauh01:
Each form error is now shown as a separate `Callout` instead of collecting all messages into a single `<ul class="errors">`. This displays the errors in a prettier way for the user.


jrauh01:
I just noticed, the callouts are from ipl-web and this is not the right place for such a feature.

PR #204, opened 2026-07-17
Make `FormElements::$populatedValues` protected

TheSyscall:
This allows the form to modify `populatedValues`


jrauh01:
Closing this as no longer needed.

For the moment we use reflection to access the value. For the next IPL HTML release we can consider adding a public getter for `$populatedValues` to not introduce a breaking change due to the visibility widening.

PR #205, opened 2026-07-21
Add `getPopulatedValues()` to `FormElements`

jrauh01:
Since `getPopulatedValue()` returns only the last entry in the populated-values chain there is no way to access previously populated values. `getPopulatedValues()` exposes the full chain, defaulting to an empty array if no value was populated.

### DIFF
--- a/src/FormElement/FormElements.php
+++ b/src/FormElement/FormElements.php
@@ -386,13 +386,14 @@ trait FormElements
     }
 
     /**
-     * Get the populated values chain of the element specified by name
+     * Get the populated values of the element specified by name
      *
-     * Returns an empty array if there is no populated value for this element.
+     * Values are returned in population order, from earliest to latest.
+     * Returns an empty array if no value was populated for this element.
      *
      * @param string $name Element name
      *
-     * @return array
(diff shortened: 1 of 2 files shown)


src/FormElement/FormElements.php
```diff
@@ -385,6 +385,20 @@ public function getPopulatedValue($name, $default = null)
...
     }
 
+    /**
+     * Get the populated values chain of the element specified by name
+     *
+     * Returns an empty array if there is no populated value for this element.
+     *
+     * @param string $name Element name
+     *
+     * @return array
```
lippserd: Make the ordering and element type explicit so callers can rely on the
complete public contract.

```suggestion
     * Get the populated values of the element specified by name
     *
     * Values are returned in population order, from earliest to latest.
     * Returns an empty array if no value was populated for this element.
     *
     * @param string $name Element name
     *
     * @return list<mixed>
```

lippserd (changes requested):
The getter is small and well covered. I left one inline request to make the public return type and ordering contract explicit.

lippserd (approved):

PR #207, opened 2026-07-30
Move `isValidEvent` into `Form` and `FieldsetElement` and add `ON_ASSEMBLED` support

jrauh01:
`Form` and `FieldsetElement` previously rejected `ON_ASSEMBLED` as an invalid event, causing `on(HtmlDocument::ON_ASSEMBLED, ...)` to throw an `InvalidArgumentException` on both classes. This moves `isValidEvent` out of the `FormElements` trait and into each concrete class so the allowed set can be scoped correctly per class.

`Form` accepts the full set of form lifecycle events plus `ON_ASSEMBLED`. `FieldsetElement` is restricted to `ON_ELEMENT_REGISTERED` and `ON_ASSEMBLED`, the only two events it actually emits. Previously the shared trait list caused `FieldsetElement` to accept events it never dispatches: `ON_SUBMIT`, `ON_SENT`, `ON_ERROR`, `ON_REQUEST`, and `ON_VALIDATE`. Registering listeners for those events would silently do nothing. They now correctly throw. This is a breaking change for any code that registered such listeners on a `FieldsetElement`.

Any future class that uses the `FormElements` trait without providing its own `isValidEvent` override will fall through to `Events::isValidEvent`, which accepts all event names unconditionally. That class must provide its own override to enforce a meaningful restriction.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -360,6 +360,18 @@ class Form extends BaseHtmlElement implements Contract\Form, Contract\FormElemen
         return $this;
     }
 
+    public function isValidEvent($event)
+    {
+        return in_array($event, [
+            Contract\Form::ON_SUBMIT,
+            Contract\Form::ON_SENT,
+            Contract\Form::ON_ERROR,
+            Contract\Form::ON_REQUEST,
+            Contract\Form::ON_VALIDATE,
+            static::ON_ELEMENT_REGISTERED,
+        ]);
+    }
+
     /**
      * Apply the form decoration
      *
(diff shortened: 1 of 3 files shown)


lippserd (commented):
Commit 1c7824edef42d8342b8e464e3d2e7b15623606d7 finally makes the Form import unused, right?

jrauh01:
No it doesn't, there's still a `@param` annotation. But since it's the only one now we could consider replacing it by fully qualified class name.

lippserd:
> No it doesn't, there's still a `@param` annotation. But since it's the only one now we could consider replacing it by fully qualified class name.

It's fine. Let's keep it as is.

lippserd (approved):

PR #208, opened 2026-08-03
Resolve `CheckboxElement`'s value and checked state on read

### LINKED ISSUE #140: Checkbox rendering depends on attribute order
I have identified an issue where the order of attributes in the configuration array for a checkbox element affects whether it is rendered as checked or unchecked. This behavior is unexpected because the order of attributes in the array should not matter.

The following code works as expected, rendering the checkbox as **checked**:
``` php
$checkbox = $this->createElement(
    'checkbox',
    'checkbox-xy',
    [
        'class'          => 'autosubmit',
        'checkedValue'   => 'a',
        'uncheckedValue' => 'b',
        'value'          => 'a',
    ]
);
```

However, the following code renders the checkbox as **unchecked**, even though it should be checked:
``` php
$checkbox = $this->createElement(
    'checkbox',
    'checkbox-xy',
    [
        'class'          => 'autosubmit',
        'value'          => 'a',
        'checkedValue'   => 'a',
        'uncheckedValue' => 'b',
    ]
);

```

jrauh01:
`CheckboxElement` derived two things eagerly in `setValue()`: it resolved a boolean value into a string, and it computed the `checked` flag. Both depend on `checkedValue`/`uncheckedValue`, which may not be set yet. `Attributes::set()` invokes setter callbacks immediately, in the order the attributes are given. So whichever attribute came first won:

```php
// renders checked
new CheckboxElement('test', ['checkedValue' => 'a', 'uncheckedValue' => 'b', 'value' => 'a']);

// renders unchecked
new CheckboxElement('test', ['value' => 'a', 'checkedValue' => 'a', 'uncheckedValue' => 'b']);
```

The case reported in #140 is not the only one. With `'checkedValue' => 'a'` and `'uncheckedValue' => 'b'`:

- `'value' => true` set before `checkedValue` stores the default `'y'` instead of `'a'`. It renders as checked either way, so the mismatch only surfaces in `getValue()` and in `hasValue()`, which then reports `false` for a visibly checked box.
- `'value' => false` set before `uncheckedValue` stores `'n'` instead of `'b'`, the same silent mismatch.
- `'value' => 'b'` is order-independent by accident: `'b'` matches neither the default `'y'` nor `'a'`, so the stale comparison happens to yield the correct unchecked state.
- `'checked' => true` never consulted the checked value at all, which left the element checked while `getValue()` returned `null`.

The boolean cases corrupt `getValue()` and `hasValue()` even where the checkbox happens to render correctly, so guarding `setCheckedValue()` alone would not have been enough.

## Changes

The value is now the only state a checkbox keeps, and everything else is derived from it on read:

- `getValue()` resolves a boolean value to the checked or unchecked value
- `isChecked()` compares `getValue()` with `getCheckedValue()`
- `setChecked()` sets the value instead of a separate flag
- `hasValue()` delegates to `isChecked()`
- the `checked` property is gone

After the change, all five scenarios above yield a single outcome across every attribute ordering.

### Changes in behavior

`setChecked()` and the `checked` attribute now set the element's value. A checkbox that is checked without having been submitted therefore has a value: `getValue()` returns the checked value instead of `null`, and a checkbox that is both required and checked passes validation instead of failing with "This field is required."

fix #140


src/FormElement/CheckboxElement.php
```diff
@@ -24,21 +21,21 @@ class CheckboxElement extends InputElement
...
     public function isChecked()
     {
-        return $this->checked;
+        return $this->getValue() === $this->getCheckedValue();
     }
 
     /**
      * Set whether the checkbox is checked
      *
+     * This sets the element's value to the checked or unchecked value respectively.
```
lippserd: The implementation stores a boolean, not the resolved string. Resolution happens lazily in `getValue()`.

```suggestion
     * This stores a boolean that {@see getValue()} resolves to the checked or unchecked value on read.
```

src/FormElement/CheckboxElement.php
```diff
@@ -89,15 +86,24 @@ public function setUncheckedValue($uncheckedValue)
         return $this;
     }
 
-    public function setValue($value)
+    /**
+     * Get the value of the element
+     *
+     * A boolean value is resolved to the checked or unchecked value here, not
+     * already when it is set. This way these values can still change afterward,
```
lippserd: "already" is redundant here.

```suggestion
     * A boolean value is resolved to the checked or unchecked value here, not
     * when it is set. This way these values can still change afterward,
```

src/FormElement/CheckboxElement.php
```diff
@@ -127,11 +133,12 @@ public function renderUnwrapped()
     }
 
     /**
-     * Determine if the checkbox is considered "checked".
+     * Determine if the checkbox is considered "checked"
+     *
      * Returns true if the current value matches the checked value, otherwise false.
      */
```
lippserd, line 141: The docblock exists but is missing the `@return` tag.

```suggestion
     * Returns true if the current value matches the checked value, otherwise false.
     *
     * @return bool
     */
```

lippserd (changes requested):

tests/FormElement/CheckboxElementTest.php
```diff
@@ -149,4 +150,95 @@ public function testCheckboxNotValidIfRequiredAndUnchecked()
...
+    {
+        $form = (new Form())
+            ->addElement('checkbox', 'test', [
+                'checkedValue'   => '1',
+                'uncheckedValue' => '0',
+                'value'          => '1'
+            ])
+            ->populate(['test' => '0']);
+
+        $this->assertFalse($form->getElement('test')->isChecked());
```
lippserd: Sorry I missed that in my previous review, `Form::getElement()` returns `FormElement` which does not declare `isChecked()`. PHPStan flags this.

```suggestion
        /** @var CheckboxElement $element */
        $element = $form->getElement('test');
        $this->assertFalse($element->isChecked());
```

src/FormElement/CheckboxElement.php
```diff
@@ -89,15 +91,24 @@ public function setUncheckedValue($uncheckedValue)
...
+     *
+     * @return mixed
+     */
+    public function getValue()
     {
+        $value = parent::getValue();
+
         if (is_bool($value)) {
-            $value = $value ? $this->getCheckedValue() : $this->getUncheckedValue();
+            return $value ? $this->getCheckedValue() : $this->getUncheckedValue();
```
lippserd, line 108: Please clear `$this->valid` in `setCheckedValue()` and `setUncheckedValue()`. These setters can now change the effective value without invalidating its cached validation result.

I reproduced this with `value => true` and a callback validator that accepts only `'y'`. Initial validation passes. Calling `setCheckedValue('a')` changes the value to `'a'`. `isValid()` still returns `true`. An explicit `validate()` call correctly rejects the new value. Before this change, the same sequence retained the validated value `'y'`.

Add regression tests that validate the checkbox before changing a mapping. The next `isValid()` call should validate the new value automatically. Test that a valid checkbox becomes invalid after the mapping changes. Also test that an invalid checkbox becomes valid.

lippserd (changes requested):

src/FormElement/CheckboxElement.php
jrauh01: `$this->valid` is now set to `null` in both methods. I've added three tests about it.

